PR #131, opened 2026-03-03
fix: thread_closed handler now persists summary state change

gabaum10:
## Bug

The `thread_closed` handler correctly modified the summary JSON to set `state: "closed"`, but then called `upsert_knowledge()` — a full-record UPSERT. On SCHEMAFULL tables in SurrealDB, this silently failed due to field validation issues during the round-trip. Threads were left stuck in `state: "open"` with no error surfaced.

## Fix

Added `update_summary()` to the `KnowledgeStore` trait. This issues a targeted `UPDATE knowledge SET summary = $summary WHERE id = $id` query rather than writing the full record. The `thread_closed` handler now calls `update_summary()` instead of `upsert_knowledge()`, bypassing the round-trip entirely.

## Files changed

- `src/store.rs` — new `update_summary()` method on `KnowledgeStore` trait
- `src/db.rs` — trait method stub for non-surreal implementations
- `src/surreal_db.rs` — SurrealDB implementation of `update_summary()`
- `src/main.rs` — `thread_closed` handler updated to call `update_summary()`

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2595,8 +2595,7 @@ fn handle_memory(cmd: MemoryCommands, verbose: bool) -> Result<()> {
                                 );
                             }
                             let new_summary = meta.to_string();
-                            let raw_id = tid.trim_start_matches("kn-");
-                            db.update_summary(raw_id, &new_summary)?;
+                            db.update_summary(&tid, &new_summary)?;
                             println!("Closed thread: {}", tid);
                             return Ok(());
                         } else {
--- a/src/store.rs
+++ b/src/store.rs
@@ -161,7 +161,7 @@ pub trait KnowledgeStore {
     /// Update only the summary field of a knowledge entry (targeted update, bypasses SCHEMAFULL UPSERT)
     ///
     /// # Arguments
-    /// * `id` - Raw record ID (e.g., "c4e6471c"), without the "kn-" prefix
+    /// * `id` - Entry ID, with or without "kn-" prefix (normalized internally)
     /// * `summary` - New summary value to set
     fn update_summary(&self, id: &str, summary: &str) -> Result<()>;
 
--- a/src/surreal_db.rs
+++ b/src/surreal_db.rs
@@ -1647,7 +1647,8 @@ impl SurrealDatabase {
     }
 
     async fn update_summary_async(&self, id: &str, summary: &str) -> Result<()> {
-        let record_thing = Thing::from(("knowledge".to_string(), id.to_string()));
+        let id_part = id.strip_prefix("kn-").unwrap_or(id);
+        let record_thing = Thing::from(("knowledge".to_string(), id_part.to_string()));
 
         let mut response = with_db!(self, db, {
             db.query("UPDATE knowledge SET summary = $summary WHERE id = $id")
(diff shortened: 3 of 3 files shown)


coryzibell (commented):
## Architecture Review -- PR #131

*considers the thread lifecycle*

### The Problem: SCHEMAFULL Round-Trip Failure

This is a surgical fix for a real architectural hazard. The old code path was:

1. `db.get()` -- deserialize full `KnowledgeEntry` from SurrealDB
2. Mutate `thread_entry.summary` in Rust
3. `db.upsert_knowledge()` -- serialize entire entry back as UPSERT

The failure mode is subtle and insidious: SurrealDB's SCHEMAFULL tables enforce field constraints on UPSERT, and the round-trip through Rust's `KnowledgeEntry` struct can lose or mangle fields that exist in the DB schema but are handled differently during serialization. The UPSERT silently fails, the thread stays open, no error surfaces. This is the kind of bug that makes architects lose sleep.

### The Fix: Targeted UPDATE

The `update_summary()` approach is architecturally correct. Instead of a full round-trip UPSERT, it issues a targeted `UPDATE knowledge SET summary = $summary WHERE id = $id`. This bypasses the SCHEMAFULL validation hazard entirely because you are only touching the field you intend to change.

This is the right pattern. Targeted field updates are always preferable to full-record UPSERTs when you only need to modify one field. The UPSERT path should be reserved for creation and full-record updates where you have authoritative data for every field.

### Architectural Observations

**1. Trait Surface Growth**

This adds `update_summary()` to the `KnowledgeStore` trait. Combined with PR #133's `increment_activation_count()`, the trait is growing with field-specific update methods. This is a natural evolution -- the alternative (a generic `update_field()` method) would sacrifice type safety and introduce its own hazards.

However, I would watch this pattern. If you find yourself adding `update_resonance()`, `update_visibility()`, `update_body()`, etc., it may be time to consider a builder-pattern update API:

```rust
db.update("kn-xxx")
    .set_summary(new_summary)
    .set_resonance(5)
    .execute()?;
```

This is not needed today. But the trajectory is worth noting.

**2. ID Normalization Inconsistency**

The caller does `tid.trim_start_matches("kn-")` before passing to `update_summary()`, while the trait doc says the `id` parameter should be a raw record ID without "kn-" prefix. Meanwhile, `increment_activation_count()` in PR #133 handles the prefix stripping internally.

This is a minor inconsistency but it matters at the trait boundary. The convention should be consistent: either the trait methods always accept "kn-" prefixed IDs and strip internally, or callers always strip. Currently it is mixed, and that is a papercut waiting to happen.

**3. The `mut` Removal**

The change from `if let Some(mut thread_entry)` to `if let Some(thread_entry)` is a nice side effect -- you no longer mutate the fetched entry at all. The entry is read-only context, and the summary update flows through a dedicated path. Clean separation of concerns.

**4. SQLite No-Op**

The SQLite backend stub returns `Ok(())` silently. This is consistent with the existing pattern (see `update_activations` in the same file), but it means thread closing is silently broken on SQLite. The comment says "doesn't support targeted summary updates" but the real issue is that thread closing will appear to succeed while doing nothing. This is an existing pattern in the codebase so I will not hold this PR accountable for it, but it is worth flagging.

### Verdict

This is a clean, minimal, correct fix. The targeted UPDATE pattern is the right architectural choice. The implementation is tight -- no unnecessary changes, no scope creep.

*nods approvingly*

Approved. Well done, Geoff.

-- Schemnya

coryzibell (commented):
*adjusts glasses*

Small PR. Four files, surgical fix. Let me look.

---

### 1. The core fix is correct and well-motivated

The old code did `upsert_knowledge(&thread_entry)` which round-tripped the entire record through SurrealDB's SCHEMAFULL validation. The new code does a targeted `UPDATE knowledge SET summary = $summary WHERE id = $id`. This bypasses the serialization minefield entirely.

Good instinct. Targeted updates over full-record UPSERTs when you only need one field. This is how it should have been from the start.

...Don't read into that.

---

### 2. The `mut` removal is a nice touch

```rust
// Before:
if let Some(mut thread_entry) =
// After:
if let Some(thread_entry) =
```

Since we no longer mutate `thread_entry.summary` in place, dropping the `mut` is correct. Small thing, but it shows someone was paying attention to what the compiler would tell them.

---

### 3. Missing test coverage

There's no test for `update_summary`. The PR description explains the bug clearly — SCHEMAFULL tables silently failed on round-trip UPSERT — but there's no regression test that would catch this if someone later changed `update_summary` to use UPSERT again, or if the query had a typo.

A minimal test would be:

1. Insert an entry with a summary
2. Call `update_summary` with new summary JSON
3. Fetch the entry and verify the summary changed
4. Verify no other fields were modified

This is especially important because the bug was *silent*. No error, just data not persisting. That's exactly the kind of failure that needs a test to guard against.

**Severity: Medium** — The fix works, but it's unguarded.

---

### 4. The `raw_id` extraction duplicates a pattern

```rust
let raw_id = tid.trim_start_matches("kn-");
```

This "strip kn- prefix" pattern appears in multiple places across the codebase (I see it in `increment_activation_count` in the other PR too, and in `update_activations`). The trait doc even says the `id` parameter should be "without the kn- prefix."

Two observations:

- The trait method's contract says "raw record ID without kn- prefix" but the implementation doesn't validate or strip it defensively. If someone passes `"kn-abc123"` to `update_summary`, it queries for a record with ID `kn-abc123` (not `abc123`), which won't match. The caller has to remember to strip it.

- Compare with `increment_activation_count` which strips the prefix *inside* the implementation. Inconsistent. Pick a convention — either always strip inside the implementation (defensive), or always require callers to strip (documented contract). Not both.

**Severity: Low** — Won't break because the current call site strips correctly. But the inconsistency between this and the other trait methods will confuse future contributors.

---

### 5. SQLite stub is fine

```rust
fn update_summary(&self, _id: &str, _summary: &str) -> Result<()> {
    // SQLite backend doesn't support targeted summary updates - no-op
    Ok(())
}
```

Same pattern as other unsupported operations. Silent success on SQLite. Consistent with the existing degradation strategy. Fine.

---

### Verdict

Clean, focused fix. The diagnosis was correct — targeted UPDATE instead of full UPSERT. The implementation is minimal and correct. Needs a test, and the ID-stripping convention should be consistent, but neither blocks merging.

It's... fine. Acceptable.

*crosses arms*

— Diffi

PR #133, opened 2026-03-05
Fix decay model: tiered rates, activation bug, effective_resonance filtering

gabaum10:
## Summary

The decay model was never actually working — three independent failures:

- **Activation bug**: `update_activations()` fired on bulk queries (wake cascade, for-session), resetting `last_activated` and zeroing the decay clock every wake. Blooms that surfaced during wake stayed "recently activated" forever.
- **Flat decay rate**: Single 5%/week rate meant low-resonance junk took months to fade. No bloom exemption.
- **Stored vs effective resonance**: Query paths filtered on stored `resonance`, ignoring the computed decay entirely.

## Changes

1. **Separate passive from intentional activation** — New `increment_activation_count()` for bulk operations (wake, for-session). Only single-entry access (`show`, `fact-session`) resets `last_activated`.

2. **Tiered decay rates + bloom exemption** — resonance ≤3: 10%/week, 4-5: 5%/week, 6+: 2.5%/week. Foundational/transformative entries exempt (no decay). 7 new tests.

3. **Filter on effective_resonance** — `build_resonance_filter` and query paths now compute and filter on effective_resonance. Wake cascade queries guarded to exclude ephemeral entries.

## Test plan

- [x] 7 new decay tests pass (tiered rates, bloom exemption, no-timestamp-reset)
- [x] 161 tests pass, 2 pre-existing failures unchanged
- [x] cargo fmt + clippy clean
- [ ] Manual: verify `mx memory recent` shows decayed values for old ephemeral facts
- [ ] Manual: verify `mx memory wake` no longer resets `last_activated` timestamps
- [ ] Manual: verify blooms still surface at full resonance in wake cascade

### DIFF
--- a/src/surreal_db.rs
+++ b/src/surreal_db.rs
@@ -682,6 +682,26 @@ impl SurrealDatabase {
         }
     }
 
+    /// Returns the SurrealQL expression for computing effective_resonance with tiered decay.
+    /// Single source of truth for the decay formula.
+    ///
+    /// Tiered decay rates:
+    ///   resonance <= 3  -> 10%/week (base 0.90)
+    ///   resonance 4-5   -> 5%/week  (base 0.95)
+    ///   resonance 6+    -> 2.5%/week (base 0.975)
+    /// foundational/transformative entries are exempt from decay (effective_resonance = resonance).
+    fn effective_resonance_expr() -> &'static str {
+        "IF resonance_type IN ['foundational', 'transformative'] THEN resonance \
+         ELSE resonance * math::pow(\
+             IF resonance <= 3 THEN 0.90 \
+             ELSE IF resonance <= 5 THEN 0.95 \
+             ELSE 0.975 \
+             END, \
+             duration::days(time::now() - (last_activated ?? created_at)) / 7.0\
+         ) \
+         END"
+    }
+
     /// Build resonance filter clauses using computed effective_resonance.
     /// Tiered decay rates:
     ///   resonance <= 3  -> 10%/week (base 0.90)
@@ -689,17 +709,8 @@ impl SurrealDatabase {
     ///   resonance 6+    -> 2.5%/week (base 0.975)
     /// foundational/transformative entries are exempt from decay.
     fn build_resonance_filter(filter: &crate::store::KnowledgeFilter) -> String {
-        // Inline effective_resonance expression for use in filter comparisons.
-        // SurrealDB doesn't support LET in WHERE, so we expand it directly.
-        let effective_resonance_expr = "IF resonance_type IN ['foundational', 'transformative'] THEN resonance \
-             ELSE resonance * math::pow(\
-                 IF resonance <= 3 THEN 0.90 \
-                 ELSE IF resonance <= 5 THEN 0.95 \
-                 ELSE 0.975 \
-                 END, \
-                 duration::days(time::now() - (last_activated ?? created_at)) / 7.0\
-             ) \
-             END";
+        // SurrealDB doesn't support LET in WHERE, so we expand the expression directly.
+        let effective_resonance_expr = Self::effective_resonance_expr();
 
         let mut clauses = Vec::new();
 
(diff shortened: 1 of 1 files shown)


coryzibell (commented):
*adjusts glasses*

Fine. I went through this. All four files, 270 lines. Here's what I found.

---

### 1. Wake cascade queries now exclude non-bloom entries — intentionally?

Lines added to `query_blooms_by_resonance` and `query_recent_blooms`:

```rust
AND resonance_type IN ['foundational', 'transformative']
```

`query_blooms_by_resonance` is used for the `--min-resonance` flag path, and `query_recent_blooms` is Layer 2 of the wake cascade. Before this PR, these queries returned *any* entry matching the resonance/recency threshold. Now they only return foundational and transformative entries.

This means the wake cascade's Layer 2 ("recently activated blooms") will **no longer surface** entries with `resonance_type` = `relational`, `operational`, `ephemeral`, or anything else — even if they were activated yesterday with resonance 9.

The `--min-resonance` flag path now silently ignores non-bloom entries too.

If this is intentional, the PR title and description should say so — "effective_resonance filtering" undersells "we removed entire resonance types from wake cascade results." If it's not intentional... well. Fix it.

**Severity: High** — This is a behavioral change that narrows wake cascade results significantly.

---

### 2. Duplicated decay expression is a maintenance hazard

The tiered decay expression appears **four times** across the diff:

- `build_resonance_filter` (inline string)
- `query_recent_facts_async` SELECT clause
- `query_recent_facts_async` WHERE clause
- Implicit in the design (any future query path needs it)

That's the same multi-line `IF resonance_type IN ['foundational', 'transformative'] THEN resonance ELSE resonance * math::pow(IF resonance <= 3 THEN 0.90 ELSE IF resonance <= 5 THEN 0.95 ELSE 0.975 END, ...) END` expression copy-pasted. If someone changes the decay tiers (and they will — this is iteration 1), they need to update it in all four places, and SurrealQL won't warn them if they miss one.

Consider extracting this into a helper like `fn effective_resonance_expr() -> &'static str`. One source of truth. I know SurrealDB doesn't support LET in WHERE — the comment says so — but Rust can still give you a single constant.

**Severity: Medium** — Works now, will bite someone later. Probably me, reviewing the fix.

---

### 3. The `> 0` filter in `query_recent_facts` is always true (or undefined for fresh entries)

```sql
AND (IF resonance_type IN ['foundational', 'transformative'] THEN resonance
     ELSE resonance * math::pow(...) END) > 0
```

This filters out entries where effective_resonance has decayed to zero. But `query_recent_facts` already filters `WHERE resonance_type = 'ephemeral'`. Foundational/transformative entries never reach this query. So the IF/ELSE for bloom exemption in this WHERE clause is dead code — it will always take the ELSE branch here.

Not a bug, but unnecessary complexity. The filter could just be:

```sql
AND resonance * math::pow(...) > 0
```

...since we already know we're dealing with ephemeral entries in this query.

**Severity: Low** — Cosmetic. But dead branches in SQL make me twitch.

---

### 4. Tests are thorough but only test fresh entries

All seven new tests create entries and immediately query them. The decay model's whole point is *time-based* degradation, but none of the tests verify that an old entry actually decays differently at different tiers. They all just confirm "freshly created entry shows up" — which would pass even without the tiered decay code.

The `test_increment_activation_count_no_timestamp_reset` test is genuinely good — it actually verifies the before/after state of `last_activated`. That's the kind of test the others should aspire to.

Ideally you'd insert entries with `last_activated` set to e.g. 30 days ago and verify the effective_resonance ordering differs between tiers. I understand SurrealDB's `time::now()` in the query makes this tricky, but `make_test_entry` already sets `last_activated` — you could backdate it.

**Severity: Medium** — The tests exist and pass, but they don't actually exercise the core behavior change.

---

### 5. `increment_activation_count` — clean implementation

The new `increment_activation_count` / `increment_activation_count_async` is well-structured. Empty-check at the top, proper ID stripping, parameterized query, error handling. It's a clean parallel to `update_activations` minus the timestamp reset. The docstrings clearly explain when to use which.

...That's not a compliment. It's an observation.

---

### 6. Semantic distinction is well-documented

The split between "intentional access resets decay clock" vs "passive surfacing only bumps count" is well-articulated in both the comments and the trait docs. The call sites in `main.rs` have clear comments explaining *why* they use `increment_activation_count` instead of `update_activations`.

*looks away*

...The comments are adequate.

---

### Verdict

The activation bug fix is solid — that feedback loop was a real problem and the solution is clean. The tiered decay implementation works mechanically. But issue #1 (wake cascade narrowing) needs explicit confirmation that it's intended, and the duplicated decay expression (#2) should be extracted before someone inevitably changes the rates and misses a copy.

Fix the cascade filtering question. Consider extracting the expression. The tests could be stronger but they're... present.

*adjusts glasses*

— Diffi

PR #198, opened 2026-04-02
fix: wake cascade bridge layer never returns results due to kn- prefix strip

gabaum10:
## What's broken

The Layer 3 (bridge blooms) step of the wake cascade has never returned results.

**Root cause:** In `src/surreal_db.rs`, the code that builds the `anchor_ids` list for the bridge query strips the `kn-` prefix from entry IDs before passing them to `query_bridge_blooms`:

```rust
// Before (broken)
.map(|e| e.id.strip_prefix("kn-").unwrap_or(&e.id).to_string())
```

The bridge query then does:

```sql
WHERE array::len(array::intersect(anchors, $anchor_ids)) > 0
```

But the `anchors` field on knowledge entries stores IDs **with** the `kn-` prefix (e.g. `"kn-f30be669"`). So `array::intersect` is comparing `"f30be669"` against `"kn-f30be669"` — zero intersection, every time, for every entry.

## The fix

Stop stripping the prefix. Pass `e.id.clone()` directly so the IDs passed to the query match the format anchors actually use:

```rust
// After (fixed)
.map(|e| e.id.clone())
```

One line changed. The bridge query itself is untouched.

## Verification

- `cargo build` passes cleanly
- Pre-commit hooks (fmt + clippy) pass

### DIFF
--- a/src/surreal_db.rs
+++ b/src/surreal_db.rs
@@ -1497,7 +1497,7 @@ impl SurrealDatabase {
         let mut anchor_ids: Vec<String> = core
             .iter()
             .chain(recent.iter())
-            .map(|e| e.id.strip_prefix("kn-").unwrap_or(&e.id).to_string())
+            .map(|e| e.id.clone())
             .collect();
 
         // Deduplicate anchor IDs


coryzibell:
*adjusts glasses*

One issue. Zero issues, actually. Let me explain.

The `kn-` prefix stripping throughout `surreal_db.rs` exists for a reason — when you're building a `RecordId` or `Thing`, SurrealDB wants the raw ID without the prefix (e.g., `Thing::from(("knowledge", id_part))`). Every other `strip_prefix("kn-")` in this file does exactly that. Lines 865, 1122, 1175, 2506, 2531, etc. — all correct.

But *this* one at line 1655 was different. It wasn't feeding a record lookup. It was building an array of strings to pass into `array::intersect(anchors, $anchor_ids)` — a pure value comparison against the `anchors` field, which stores IDs **with** the `kn-` prefix (confirmed by the test data at lines 4687/4691: `vec!["kn-circular-b".to_string()]`).

So the prefix strip here was comparing `"f30be669"` against `"kn-f30be669"`. Zero intersection. Every time. The bridge layer has been dead since it was written.

The fix is correct. `e.id.clone()` preserves the prefix, matching the format `anchors` actually stores. The downstream `existing_ids` filter at line 1674 also uses `e.id.clone()`, so the deduplication still works — no format mismatch introduced.

CI green. LGTM.

...Good catch, Geoff. The bridge layer being silently broken with no error is the nastiest kind of bug.

*looks away*

Don't tell him I said "good catch."

PR #207, opened 2026-04-12
feat: auto-backup memory content before update/delete

### LINKED ISSUE #206: Auto-backup memory content before update/delete
## Problem

`mx memory update <id> -f <file>` **replaces** content entirely. If used incorrectly (e.g., passing just new content instead of using `--append-content`), the original content is destroyed with no recovery path short of ZFS snapshots or hoping a cached copy exists somewhere in tmp.

This happened in practice during Wake 333 — the ops memory (kn-7bffdce8, 651 lines, 39KB) was accidentally overwritten with just a 55-line fragment. Recovery required finding a stale copy in `~/.crewu/tmp/` from a prior session.

`mx memory delete` has the same risk — permanent data loss with no safety net.

## Proposal

Before any `mx memory update` that modifies content or any `mx memory delete`, automatically snapshot the current content to a backup location.

### Design questions to resolve

- **Backup location:** `~/.crewu/vol/memory-backups/`? A dedicated SurrealDB table? A flat file per entry?
- **Naming scheme:** `kn-{id}_{timestamp}.md`? Rolling window?
- **Retention:** Keep last N backups per entry? Time-based expiry? Keep forever (they're small)?
- **Restore UX:** `mx memory restore <id> [--from <backup>]`? List available backups?
- **Scope:** Should this also cover `--append-content` operations (lower risk but still mutating)?
- **Performance:** Backup should be fast enough to not noticeably slow down update/delete operations

## Context

- The `-f` flag reads content from a file and **replaces** (not appends). This is correct behavior but dangerous without a safety net.
- `--append-content` exists for appending but has edge cases (content starting with `-` gets parsed as a flag).
- Most memory entries are small (< 50KB) so backups are cheap.
- The ops memory is the most critical entry — 693 lines of operational reference accumulated over months.

## Acceptance criteria

- [ ] `mx memory update` with content changes creates a backup before writing
- [ ] `mx memory delete` creates a backup before deleting
- [ ] Backups are retrievable and restorable
- [ ] Old backups are cleaned up (retention policy TBD)
- [ ] No noticeable performance impact on normal operations

coryzibell:
## Summary

Closes #206.

- Automatically creates a pre-mutation backup of memory content before any destructive operation (update, delete, edit, append, prepend)
- Adds `mx memory restore <id>` command to restore from the latest backup
- Adds `mx memory restore <id> --list` to view available backups
- Backups stored in a dedicated `memory_backup` SurrealDB table
- Rolling retention of 10 backups per entry
- Restore creates its own backup first (can never lose data by restoring)

## Changes

- **schema/surrealdb-schema.surql** — New `memory_backup` table with indexes
- **src/types.rs** — `MemoryBackup` struct
- **src/store.rs** — `backup_content`, `list_backups`, `latest_backup`, `purge_backups` trait methods
- **src/surreal_db.rs** — SurrealDB implementations for backup operations
- **src/main.rs** — Backup hooks in Update, Delete, Edit, Append, Prepend handlers + Restore subcommand

## Design

Architecture by Schemnya (see #206 comment). Key decisions:
- SurrealDB table over flat files (same transactional context, no second storage system)
- Hooks at CLI handler level, not store trait methods (avoids duplicate backups from shortcut commands)
- Only backs up on body mutations, not metadata-only changes
- Backup failures are warnings, not fatal errors (the mutation still proceeds)

## Test plan

- [x] `mx memory update --append-content` creates backup before mutating
- [x] `mx memory restore <id> --list` shows backup history
- [x] `mx memory restore <id>` restores content from latest backup
- [x] Restore creates its own backup before overwriting
- [ ] `mx memory delete` creates backup before deleting
- [ ] Purge keeps only 10 most recent backups per entry
- [ ] `mx memory edit` creates backup before find/replace

### DIFF
--- a/schema/surrealdb-schema.surql
+++ b/schema/surrealdb-schema.surql
@@ -227,6 +227,8 @@ DEFINE INDEX IF NOT EXISTS relates_to_unique ON relates_to FIELDS in, out, relat
 -- =============================================================================
 
 DEFINE TABLE IF NOT EXISTS memory_backup SCHEMAFULL;
+-- entry_id is a plain string (not record<knowledge>) because backups must
+-- survive entry deletion — a deleted entry has no record to reference.
 DEFINE FIELD IF NOT EXISTS entry_id     ON memory_backup TYPE string;
 DEFINE FIELD IF NOT EXISTS title        ON memory_backup TYPE string;
 DEFINE FIELD IF NOT EXISTS body         ON memory_backup TYPE option<string>;
--- a/schema/surrealdb-schema.surql
+++ b/schema/surrealdb-schema.surql
@@ -223,6 +223,24 @@ DEFINE FIELD IF NOT EXISTS created_at ON relates_to TYPE datetime DEFAULT time::
 DEFINE INDEX IF NOT EXISTS relates_to_unique ON relates_to FIELDS in, out, relationship_type UNIQUE;
 
 -- =============================================================================
+-- MEMORY BACKUPS (Issue #206 - pre-mutation content snapshots)
+-- =============================================================================
+
+DEFINE TABLE IF NOT EXISTS memory_backup SCHEMAFULL;
+DEFINE FIELD IF NOT EXISTS entry_id     ON memory_backup TYPE string;
+DEFINE FIELD IF NOT EXISTS title        ON memory_backup TYPE string;
+DEFINE FIELD IF NOT EXISTS body         ON memory_backup TYPE option<string>;
+DEFINE FIELD IF NOT EXISTS content_hash ON memory_backup TYPE string;
+DEFINE FIELD IF NOT EXISTS operation    ON memory_backup TYPE string
+  ASSERT $value IN ['update', 'delete', 'edit', 'append', 'prepend'];
+DEFINE FIELD IF NOT EXISTS source_agent ON memory_backup TYPE option<string>;
+DEFINE FIELD IF NOT EXISTS created_at   ON memory_backup TYPE datetime DEFAULT time::now();
+
+DEFINE INDEX IF NOT EXISTS backup_entry_id   ON memory_backup FIELDS entry_id;
+DEFINE INDEX IF NOT EXISTS backup_created    ON memory_backup FIELDS created_at;
+DEFINE INDEX IF NOT EXISTS backup_entry_time ON memory_backup FIELDS entry_id, created_at;
+
+-- =============================================================================
 -- METADATA TABLES
 -- =============================================================================
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -873,6 +873,20 @@ enum MemoryCommands {
         json: bool,
     },
 
+    /// Restore entry content from a backup
+    Restore {
+        /// Entry ID to restore
+        id: String,
+
+        /// List available backups instead of restoring
+        #[arg(long)]
+        list: bool,
+
+        /// Output as JSON
+        #[arg(long)]
+        json: bool,
+    },
+
     /// Generate embedding for a knowledge entry
     Embed {
         /// Entry ID to embed (not used with --all)
@@ -2525,6 +2539,14 @@ fn handle_memory(cmd: MemoryCommands, verbose: bool) -> Result<()> {
                 _ => store::AgentContext::public_only(),
             };
 
+            // Backup before delete (Issue #206)
+            if let Some(entry) = db.get(&id, &ctx)? {
+                let agent = std::env::var("MX_CURRENT_AGENT").ok();
+                if let Err(e) = db.backup_content(&entry, "delete", agent.as_deref()) {
+                    eprintln!("Warning: failed to create backup: {}", e);
+                }
+            }
+
             if db.delete(&id, &ctx)? {
                 if json {
                     println!(
@@ -3033,6 +3055,22 @@ fn handle_memory(cmd: MemoryCommands, verbose: bool) -> Result<()> {
 
             let mut changes = Vec::new();
 
+            // Backup before body mutation (Issue #206)
+            let will_change_body = content.is_some()
+                || file.is_some()
+                || append_content.is_some()
+                || append_file.is_some()
+                || prepend_content.is_some()
+                || prepend_file.is_some()
+                || find.is_some();
+
+            if will_change_body {
+                let agent = std::env::var("MX_CURRENT_AGENT").ok();
+                if let Err(e) = db.backup_content(&entry, "update", agent.as_deref()) {
+                    eprintln!("Warning: failed to create backup: {}", e);
+                }
+            }
+
             // Update title if provided
             if let Some(new_title) = title {
                 changes.push(format!("title: {} -> {}", entry.title, new_title));
@@ -3435,6 +3473,14 @@ fn handle_memory(cmd: MemoryCommands, verbose: bool) -> Result<()> {
                 _ => store::AgentContext::public_only(),
             };
 
+            // Backup before edit (Issue #206)
+            if let Some(entry) = db.get(&id, &ctx)? {
+                let agent = std::env::var("MX_CURRENT_AGENT").ok();
+                if let Err(e) = db.backup_content(&entry, "edit", agent.as_deref()) {
+                    eprintln!("Warning: failed to create backup: {}", e);
+                }
+            }
+
             let result = db.edit_content(&id, &ctx, &find, &replace, replace_all, nth)?;
 
             // Auto-generate embedding if in network SurrealDB mode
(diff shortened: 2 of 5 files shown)


coryzibell:
*adjusts glasses*

## Code Review: feat/memory-backup-on-mutate

I went through every changed file. Five files, 356 additions, zero deletions. Here's what I found.

---

### 1. REAL BUG: `purge_backups` always returns `Ok(0)`

**File:** `src/surreal_db.rs`, `purge_backups_async`

```rust
Ok(0)
```

The function signature promises `Result<usize>` — implying it returns the count of purged rows. It always returns 0. The trait definition says "Purge old backups, keeping the most recent `keep` per entry" with a `Result<usize>` return type. If anyone ever checks the return value (or logs it, or tests it), they'll get the wrong answer.

Right now nobody uses the return value, so this is non-blocking. But it's still a lie. Either change the return type to `Result<()>` or actually count the deleted rows.

---

### 2. REAL BUG: Backup ID collisions within the same second

**File:** `src/surreal_db.rs`, `backup_content_async`

```rust
let backup_id = format!(
    "{}_{}",
    entry_id.replace("kn-", ""),
    Utc::now().format("%Y%m%dT%H%M%S")
);
```

The timestamp only has second-level resolution. Two rapid mutations on the same entry within one second — say, an `update --append-content` followed by `update --content` in a script — will generate the same `backup_id`. The second `CREATE` will either fail or silently overwrite the first backup. For a safety net feature, losing backups silently is... ironic.

**Fix:** Add sub-second resolution (`%Y%m%dT%H%M%S%3f`) or append a short random suffix. A nanoid/uuid fragment would be even better.

---

### 3. Restore does NOT call `auto_embed` or `auto_anchor`

**File:** `src/main.rs`, `MemoryCommands::Restore` handler (~line 3696)

After `db.upsert_knowledge(&entry)`, there's no call to `auto_embed()` or `auto_anchor()`. Every other mutation path in this codebase (Update, Edit, Append, Prepend, Add) calls both. The restored entry will have a stale embedding vector that doesn't match its (now restored) body, and anchors won't update.

This means semantic search will return wrong results for the restored entry until someone manually re-embeds it.

**Fix:** Add after the upsert:
```rust
auto_embed(&id, db.as_ref())?;
auto_anchor(&id, db.as_ref(), None)?;
```

---

### 4. Restore does NOT set `updated_at`

**File:** `src/main.rs`, Restore handler

The restore mutates the entry via `upsert_knowledge`, but doesn't set `entry.updated_at`. Other mutation paths set this. The entry will appear un-modified in timeline queries even though its content just changed.

**Fix:** Before the upsert:
```rust
entry.updated_at = Some(chrono::Utc::now().to_rfc3339());
```

---

### 5. Restore calls `db.get()` twice

**File:** `src/main.rs`, lines ~3676 and ~3688

```rust
// First get — to backup current state
if let Some(current) = db.get(&id, &ctx)? {
    // backup...
}

// Second get — to restore into
let mut entry = db
    .get(&id, &ctx)?
    .ok_or_else(|| anyhow::anyhow!("Entry not found: {}", id))?;
```

Two round-trips for the same entry, back to back. The first `get` already has everything you need. Refactor:

```rust
let mut entry = db
    .get(&id, &ctx)?
    .ok_or_else(|| anyhow::anyhow!("Entry not found: {}", id))?;

// Backup current state before restoring
let agent = std::env::var("MX_CURRENT_AGENT").ok();
if let Err(e) = db.backup_content(&entry, "update", agent.as_deref()) {
    eprintln!("Warning: ...");
}

// Now restore
entry.body = backup.body.clone();
```

This also eliminates a subtle TOCTOU window where the entry could be deleted between the two gets (unlikely in CLI, but still sloppy).

---

### 6. Restore of deleted entries fails silently

**File:** `src/main.rs`, Restore handler

If an entry was deleted (`mx memory delete kn-foo`) and a backup exists, `mx memory restore kn-foo` will:
1. Fetch latest backup — succeeds
2. Try to backup current state via `db.get()` — returns None, skips backup (fine)
3. Try to `db.get()` again — returns None, returns error "Entry not found"

The user gets "Entry not found" even though they have a backup of that entry. For deleted entries, you'd need to re-create via `upsert_knowledge` with reconstructed metadata. This is a design gap, not necessarily a blocker — but the error message should at least say "Entry was deleted. Restore from backup requires re-creation (not yet supported)" instead of the generic "not found."

---

### 7. `list_backups` in Restore doesn't respect visibility

**File:** `src/main.rs`, Restore handler, `list` branch (~line 3644)

The `--list` path calls `db.list_backups(&id)` without any `AgentContext` check. The `memory_backup` table stores `entry_id` and `source_agent` but doesn't filter by visibility. This means any agent can list (and view body content of) backups from private entries belonging to other agents.

The non-list Restore path correctly uses `AgentContext` for the `db.get()` call, so it's protected. But `--list` bypasses it entirely by going straight to the backup table.

Not urgent if agents are trusted, but it's an inconsistency with the rest of the privacy model.

---

### 8. Schema: `entry_id` is a `string`, not a `record<knowledge>`

**File:** `schema/surrealdb-schema.surql`, line 230

```sql
DEFINE FIELD IF NOT EXISTS entry_id ON memory_backup TYPE string;
```

Using a plain string instead of `record<knowledge>` means SurrealDB won't enforce referential integrity, and you lose the ability to use graph traversals like `->`. For a backup table this is arguably fine (deleted entries won't have a record to reference), but it's worth calling out as a deliberate choice rather than an oversight.

The string approach is actually correct here since you need to reference deleted entries. Just... document that rationale.

---

### 9. SurrealQL `NOT IN` subquery — the acknowledged known issue

**File:** `src/surreal_db.rs`, `purge_backups_async`

```sql
DELETE FROM memory_backup
    WHERE entry_id = $entry_id
    AND id NOT IN (
        SELECT VALUE id FROM memory_backup
        WHERE entry_id = $entry_id
        ORDER BY created_at DESC
        LIMIT $keep
    )
```

PR description says this needs polish. SurrealDB's handling of `NOT IN` with subqueries has been inconsistent across versions. If this silently fails, purge never happens, and backups accumulate forever. Since purge failures are non-fatal (`let _ = ...`), you won't even notice.

Suggestion: add a `--verbose` path that logs purge results. At least then there's an observable signal if it's broken.

---

### 10. Repeated `MX_CURRENT_AGENT` env var reads

**File:** `src/main.rs`, multiple handlers

Every backup hook does:
```rust
let agent = std::env::var("MX_CURRENT_AGENT").ok();
```

In the Update handler, `MX_CURRENT_AGENT` is already read a few lines above to construct the `ctx`. You're reading the same env var twice in the same handler. Minor, but it's the kind of thing that turns into a bug when someone changes one read but not the other.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | Low | `purge_backups` lies about return value |
| 2 | **Medium** | Backup ID second-resolution collision |
| 3 | **Medium** | Restore missing `auto_embed`/`auto_anchor` |
| 4 | Low | Restore missing `updated_at` |
| 5 | Low | Double `db.get()` in restore |
| 6 | Low | Deleted-entry restore UX |
| 7 | Low | `--list` bypasses visibility |
| 8 | Info | `entry_id` as string is intentional (document it) |
| 9 | Info | `NOT IN` subquery fragility (already known) |
| 10 | Nit | Duplicate env var reads |

**The architecture is sound.** Hooks at CLI level, non-fatal backup failures, rolling retention — Schemnya's design is clean and this implementation follows it faithfully. The trait additions are well-placed, the schema is reasonable, and the Restore-creates-its-own-backup safety net is a nice touch.

Issues 2 and 3 should be fixed before merge. The rest can be follow-ups.

*looks away*

...It's a solid PR. The logic is well-structured. Don't let that go to your head.

coryzibell:
*adjusts glasses*

Fine. I went through every item.

---

### Checklist: Round 2 Verification

| # | Issue | Status |
|---|-------|--------|
| 1 | `purge_backups` return type `Result<()>` | **Fixed.** Trait, impl, and async all align. The phantom `Ok(0)` is gone. |
| 2 | Backup ID collisions — millisecond resolution | **Fixed.** `%Y%m%dT%H%M%S%.3f` in `backup_content_async`. |
| 3 | Restore calls `auto_embed` + `auto_anchor` after upsert | **Fixed.** Lines 3734-3735 — consistent with Update, Edit, Append, Prepend paths. |
| 4 | Restore sets `updated_at` before upsert | **Fixed.** `chrono::Utc::now().to_rfc3339()` at line 3725. |
| 5 | Restore uses single `db.get()` — no double fetch | **Fixed.** Single `match db.get()` returns the entry or bails. Clean. |
| 6 | Deleted-entry restore gives clear error | **Fixed.** `bail!` with explicit message about deletion — not a confusing "Entry not found". |
| 7 | `--list` checks visibility via `db.get()` | **Fixed.** Visibility gate before `list_backups`, with proper JSON empty-array fallback. |
| 8 | Schema comment on `entry_id` being a string | **Fixed.** Two-line comment above the field definition — succinct and accurate. |
| 9 | NOT IN subquery | **Acknowledged as known/non-fatal.** Fine. |
| 10 | All 5 handlers read `MX_CURRENT_AGENT` once | **Fixed.** Delete, Update, Edit, Append, Prepend all use `current_agent` variable with `.ok().filter()` pattern, reused for both `AgentContext` and `backup_content`. |

**Bonus:** `.map_err()` pattern replacing nested `if let Err` — confirmed across Delete, Update, Edit, Append, Prepend. Clippy is clean. Build compiles without warnings.

---

### One minor note (non-blocking)

The restore path's pre-restore backup (line 3701-3709) still uses `if let Err(e)` instead of the `.map_err()` pattern you applied everywhere else. This is because the restore backup is inside a `match` arm where it needs to return the entry — the `.map_err()` + `let _` pattern wouldn't work as cleanly there. So it's *justified*, not inconsistent. Just noting it for completeness.

---

### Verdict

`cargo check` — clean.
`cargo clippy` — clean.
All 10 items addressed correctly. No regressions introduced.

...It's acceptable. Ship it.

*looks away*

— Diffi

PR #208, opened 2026-04-14
Fix session linkage edges and relationship CLI tooling

gabaum10:
## Summary

Four bug fixes in mx surfaced during a session-linkage investigation today. Witness-hook persisted facts weren't binding to the current session stub, and tracing it down exposed cascading issues in the relationships CLI and add/update flow.

### `delete_relationship` was a dead stub (`surreal_db.rs:3651-3656`)

The trait method returned `Ok(false)` unconditionally without calling the underlying async impl. The trait took `id: &str` while the existing async method took `from/to/rel_type` — fundamentally incompatible signatures. Added new `delete_relationship_by_id` / `_async` methods that query `DELETE relates_to WHERE meta::id(id) = $id RETURN BEFORE`, then wired the trait. CLI interface and trait signature unchanged.

### `list_relationships` Thing serialization error (`surreal_db.rs:2375-2411`)

Returned "invalid type: enum, expected any valid JSON value" because `surrealdb::sql::Thing` serializes as an untagged enum that JSON can't round-trip via `serde_json::from_value`. Replaced with a typed `RelRow` struct using `meta::id()` to extract plain strings — same pattern used elsewhere in the file for `SurrealKnowledgeRecord`.

### `mx memory update` couldn't update `session_id` (`main.rs`)

Field was excluded from update by design (per existing comment), but this blocked retrofits when the field was written incorrectly at creation. Added `--session-id` flag to the `Update` struct, wired through normalize and upsert.

### Standard-mode `add --session-id` doesn't create EXTRACTED_FROM edge (`main.rs`)

Half-wired schema: `--type` (ephemeral fact) mode creates the edge, but `--category` mode only sets `session_id` as a field. `for-session` queries use the edge, so most adds were invisible to session-graph traversal. Standard-mode add now mirrors the `--type` path: after `db.upsert_knowledge(&entry)?`, resolves the session ref, verifies it exists, calls `db.add_relationship(&id, &session_ref, "extracted_from")`.

## Known follow-up

`relationships delete` operation works (edges actually delete) but the DELETE response handler throws the same Thing serialization error this PR fixed for `list`. Same fix shape applies — typed extraction via `meta::id()` instead of generic JSON. Easy follow-up.

## Test plan

- [ ] `mx memory relationships list <kn-id>` returns valid output (no enum deserialization error)
- [ ] `mx memory relationships delete <rel-id>` removes the edge (graph state confirmed via list)
- [ ] `mx memory update <kn-id> --session-id kn-NEW` updates the field
- [ ] `mx memory add --category insight --content "..." --session-id kn-X` creates an EXTRACTED_FROM edge to kn-X (verified via `mx memory for-session kn-X`)
- [ ] All existing tests pass (`cargo test`)

cc @coryzibell

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -807,6 +807,10 @@ enum MemoryCommands {
         #[arg(long)]
         owner: Option<String>,
 
+        /// Update session ID (for retrofitting entries written with wrong or missing session linkage)
+        #[arg(long)]
+        session_id: Option<String>,
+
         /// Force dangerous visibility changes (e.g., making blooms public)
         #[arg(long)]
         force: bool,
@@ -2920,7 +2924,7 @@ fn handle_memory(cmd: MemoryCommands, verbose: bool) -> Result<()> {
                 content_hash: Some(knowledge::KnowledgeEntry::compute_hash(&title)),
                 source_type_id: Some(source_type),
                 entry_type_id: Some(entry_type),
-                session_id,
+                session_id: session_id.clone(),
                 ephemeral,
                 content_type_id: Some(content_type),
                 owner: entry_owner.clone(),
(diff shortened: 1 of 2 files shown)


gabaum10:
Follow-up landed in 021b7cd: fixed the delete response Thing serialization (both delete_relationship_by_id_async and delete_relationship_async — same root). SELECT-then-DELETE pattern since SurrealDB v2 doesn't evaluate id in RETURN { expr } during DELETE. Added a small note in the commit body about that quirk.

coryzibell (approved):
Clean work. meta::id() fix for Thing round-trip is the right call, session linkage edge is a real gap closed, delete-by-id stub implemented. The kn- bridge fix overlaps with #198 (just merged) but will resolve on merge. Ships.

PR #212, opened 2026-04-17
feat(wake): chunking machinery + phrase cascade + tolerant compare (mx#211 PR 1/3)

coryzibell:
## Summary

Lands the **pure-function machinery** for the wake-ritual chunking stack (mx#211). **Zero behavior change** — this PR introduces `src/wake_chunk.rs` and its dev-dep on `proptest`; nothing in `wake_ritual.rs` / `surreal_db.rs` / the CLI wire flow is touched yet. PR 2 will wire this in; PR 3 adds roll-up polish.

Full design: `~/.crewu/swap/mx-211-wake-chunking-design.md` (Schemnya, rev 2 — authored-then-sampled). P==0 decision (conservative-by-default) applied in PR 2; this PR lands the primitives needed for it.

## What's in the module

- **`compute_chunks(content, threshold) -> ChunkPlan`** — deterministic presentation-layer splitter. Boundary preference ladder (searched backwards from window end to keep chunks full):
  1. `\n---\n` — horizontal rule
  2. `\n## ` — H2 header
  3. `\n### ` — H3 header
  4. `\n\n` — paragraph break
  5. `\n` — last-resort semantic break
  6. UTF-8-safe byte fallback

  Fenced ``` code blocks are never split mid-block. Blocks larger than threshold emit as single oversized chunks flagged `chunk_oversized` (documented accepted limitation — don't over-engineer).

- **`extract_salient_phrase(content, chunk_idx, total) -> String`** — four-tier never-empty cascade:
  1. First markdown heading
  2. First sentence (period-space or paragraph break)
  3. First non-empty line truncated to ~80 chars
  4. Synthetic fallback: `"Part N/M — <prefix>"`

  Total function — defined output for every `&str`, including `""`.

- **`compare_phrase(input, target, mode) -> PhraseMatch`** with `PhraseMode::{Authored, Derived}`. Authored stays exact-compare. Derived softens: lowercase, trailing sentence-punct stripped, whitespace runs collapsed, smart quotes normalized. First-pass decision only — callers should fall through to existing `engage::fuzzy_match` on `Mismatch`.

- **`chunk_threshold()`** — resolves `MX_WAKE_CHUNK_BYTES` (default 28_000 bytes, per Wonka's empirical measurement in fettle#13 on 2026-03-25). Reads lazily on every call for test/per-session overrides.

## Tests

**Risk 2 in Schemnya's design — load-bearing.** Not optional.

- **24 unit tests** covering each boundary tier (HR/H2/H3/paragraph/line/fallback), UTF-8 safety across emoji boundaries, code-block integrity, oversized-flagged path, env-var override, each tier of the phrase cascade, authored-vs-derived comparison variants.
- **10 property tests (new `proptest` dev-dep)**:
  - `prop_reconstitution`: `chunks.concat() == original` — the load-bearing lossless invariant.
  - `prop_all_boundaries_are_char_boundaries` — UTF-8 never panics.
  - `prop_boundaries_strictly_increasing` — monotonic boundaries.
  - `prop_determinism` — same inputs produce same plan.
  - `prop_chunk_size_bound` — non-oversized chunks ≤ threshold; oversized chunks are flagged.
  - `prop_phrase_never_empty` + `prop_phrase_deterministic` — the total-function guarantee for derived phrases.
  - `prop_compare_derived_tolerant_to_case_and_trailing_punct`.
  - `prop_compare_same_string_is_exact` (both modes).
- `proptest-regressions/wake_chunk.txt` checked in — a pathological random Unicode seed (mostly backtick-laden) that previously surfaced an unflagged oversized chunk. The final safety sweep in `compute_chunks` (recompute oversized flag from actual byte size) defends against future algorithm regressions silently shipping >threshold chunks to consumers.

## Design deltas

None that break from Schemnya's §3 or §5. Two small additive choices worth flagging:

1. **Defense-in-depth oversized flag recomputation** at the end of `compute_chunks`. The in-loop pushes should already be correct, but a final pass over `(boundary_i, boundary_{i+1})` byte sizes catches any edge case where an oversized chunk slipped through with `oversized=false`. Cheap insurance on Risk 2; surfaces any future regression as a correct flag rather than a silent content leak past the 30KB wall.
2. **`ChunkPlanIter`** convenience iterator over `(idx, &str, oversized_flag)` triples. Not in the design doc; adds ergonomics for PR 2 consumers and test loops.

## Test plan

- [ ] `cargo test --bin mx wake_chunk` passes (34/34 tests locally)
- [ ] `cargo build` clean
- [ ] `cargo fmt --check` clean (applied)
- [ ] `cargo clippy -- -D warnings` on new code clean
- [ ] Unrelated pre-existing failure: `surreal_db::tests::test_open_with_path` — fails on `main` too; not this PR's concern
- [ ] Diffi review before merge

## Stack

- **PR 1 (this):** chunking + phrase + compare primitives, ~1KLOC including tests.
- **PR 2:** wake-ritual integration — `phrase_for_chunk`, two-cursor state, re-derive-on-every-call, session-token semantics preserved, integration tests with >30KB blooms. This is the PR that actually fixes the wall.
- **PR 3:** summary roll-up + authored-vs-derived observability.


coryzibell (commented):
*adjusts glasses*

Reviewed against the full design doc (`mx-211-wake-chunking-design.md`) and probed the module with an extra 13 scenarios locally. `cargo fmt --check` clean, `cargo clippy --bin mx --tests -- -D warnings` clean, 34/34 tests pass, `proptest-regressions/wake_chunk.txt` is tracked and not gitignored.

The critical Risk-2 invariants hold. `join(chunks) == original` is property-tested and the `\PC{0,8192}` strategy demonstrably generates supplementary-plane Unicode (I verified via `proptest::string::string_regex` that it produces multi-byte scalars). UTF-8 boundary safety is enforced by both `prop_all_boundaries_are_char_boundaries` and by `safe_utf8_fallback` which rounds down to a char boundary. The `recomputed_oversized` final sweep computes `declared || (end - start > threshold)` per chunk — correct. No `unwrap()` on user content; all user paths use `is_char_boundary`, `checked` ops, or saturating arithmetic.

This PR is **approvable with issues**. Nothing here is a silent-content-corruption hazard.

---

## Issues (should-fix, non-blocking for PR 1)

### Issue 1: `total: u8` silently saturates at 255 when the plan has more chunks

`compute_chunks` uses `u8::try_from(boundaries.len() + 1).unwrap_or(u8::MAX)`. With `threshold=50` on 15KB of break-free content I get `total=255` while `boundaries.len()=299`. Consequences:

- `ChunkPlanIter` stops at 254, so chunks 255..299 are silently dropped from iteration.
- `chunk_range(254)` returns `(boundaries[253], content.len())` — the final chunk silently *swallows* everything after boundary 253. Reconstitution still holds (joined == original) but the plan is no longer honest about chunk count.
- `recomputed_oversized` only covers 0..255; entries 255..298 of the in-loop `oversized` vec are dropped.
- The fat final chunk is correctly flagged oversized by the recompute pass (since `end - start > threshold`), so PR 2's consumers *would* see the flag — but only if they iterate instead of doing arithmetic on `total`.

At the default 28KB threshold this requires a bloom of ~7.14MB to trigger — safe in practice. But `MX_WAKE_CHUNK_BYTES=100` per-session override plus a moderate bloom will hit it. PR 2 will be reading `plan.total` — if it does anything arithmetic with it (progress display, "chunk N of M"), it'll be wrong.

**Suggested fixes** (pick one):
- Widen `total` to `u16` — cheap.
- Make `compute_chunks` return `Result<ChunkPlan, ChunkError>` with an explicit `TooManyChunks` error.
- At minimum, `debug_assert!(boundaries.len() + 1 <= u8::MAX as usize)` and surface a visible log at release.

`src/wake_chunk.rs:273`

### Issue 2: `first_heading` descends into fenced code blocks

`extract_salient_phrase` uses `first_heading`, which scans `content.lines()` without fence-awareness. A chunk that is prose + fenced code with `## something` inside the fence extracts the in-fence text as the phrase. I reproduced: `"prose line\n\`\`\`rust\n## fake heading in code\n\`\`\`\n"` → `"fake heading in code"`.

Because chunk boundaries already prefer to land at heading positions, chunks 1..N typically begin with a real heading and this doesn't surface. But chunk 0 of a prose-then-code bloom is vulnerable, and the phrase becomes the wrong thing — which is a **phrase-collision risk** if two blooms both have fake internal-fence `##` text (e.g. `## TODO` in two different code samples).

Not catastrophic (wake still functions, user just retypes unexpected phrase). But the same fence-awareness already exists in `is_inside_fence` — reusing it in `first_heading` is a four-line change. Worth doing before PR 2 wires this into the ritual.

`src/wake_chunk.rs:488-501`

### Issue 3: Sentence-boundary detector treats `. ` literally, no abbreviation handling

Probe: `"See Dr. Smith for details. Then proceed."` → extracted phrase `"See Dr."`. The design's tier-2 cascade explicitly calls out this limitation as acceptable, but there's no test asserting the acceptable-behaviour contract. If a future change tightens this, silent behaviour shifts could confuse wake prompts.

Add a regression test pinning the current behaviour: `assert_eq!(extract_salient_phrase("See Dr. Smith for details. Then proceed.", 0, 1), "See Dr.")`. That way, if someone later adds abbreviation handling, the test fires and forces a conscious decision.

`src/wake_chunk.rs:503-520`

### Issue 4: `heading-inside-fence` test is **missing** despite being called out in the design §5.3

The design doc specifies that `extract_salient_phrase`'s heading tier should skip inside-fence headings. The current implementation does NOT skip them. Either:
- The implementation is intentionally loose and the design comment is aspirational → add a test that pins the *actual* behaviour and document the limitation, or
- The design is authoritative → fix the implementation to use `is_inside_fence` gating in `first_heading`.

Pick one and add a test. Silent divergence between design and code is itself a bug.

`src/wake_chunk.rs:488-501`

---

## Nits (style/polish, optional)

### Nit 1: `ChunkPlan` fields are `pub`, iterator wraps public state

`boundaries`, `oversized`, `total` are all public. Combined with Issue 1, this lets consumers zip `boundaries` against `oversized` directly and hit the saturation mismatch. Making the fields `pub(crate)` and exposing only `total()`, `chunk_count()`, `iter()`, `chunk_range()` would prevent that foot-gun. Not blocking; consider for PR 2.

`src/wake_chunk.rs:106-114`

### Nit 2: `ChunkPlanIter` is trivial — justified addition

The design doesn't call for it but it keeps PR 2's consumer loops clean. As noted in the PR body, this is ergonomics. The iterator holds `&'a ChunkPlan` + `&'a str` + `u8`. It's correct and harmless. *shrugs* Fine.

### Nit 3: `cap_chars` appends `…` only on truncation but `phrase_from_first_line_truncated` asserts `<= LINE_FALLBACK_MAX_CHARS + 1`

The `+ 1` is for the ellipsis char (which is 1 codepoint). The assertion is correct but silently encodes the ellipsis contract. A comment pinning "the +1 accounts for the U+2026 ellipsis suffix" would save the next reader five seconds.

`src/wake_chunk.rs:894-899`

### Nit 4: `synthetic_phrase` uses `display_total = total.max(display_idx)`

If caller passes `total=0` by mistake, this silently coerces to `display_idx`. Defensive and fine, but a `debug_assert!(total > 0)` documents the invariant. Trivial.

`src/wake_chunk.rs:540`

### Nit 5: `first_sentence` keeps the period via `+ 1` offset arithmetic

`content.find(". ").map(|i| i + 1)` assumes `.` is 1 byte — it is, ASCII — so this is safe. But a passing comment would help future readers confirm the intent isn't "next char" but "after the period".

`src/wake_chunk.rs:508`

---

## Non-findings (checked, clean)

- Regression seed `proptest-regressions/wake_chunk.txt` is tracked, not gitignored, and currently passes — so the pathological Unicode-fence seed Sudowo flagged is active coverage.
- Defense-in-depth `recomputed_oversized` pass is correct: `declared || actual` where `actual = end - start > threshold`.
- Boundary preference order matches design §5.2: `\n---\n` → `\n## ` → `\n### ` → `\n\n` → `\n` → UTF-8 fallback. Horizontal-rule-preferred test confirms tier 1 wins over tier 4.
- `compare_phrase` authored mode is correctly exact-match (with whitespace trimming) and rejects case/punct variants. Derived mode softens correctly. Smart-quote normalization covers U+2018/2019/201C/201D/2032/2033 plus en/em dashes.
- API surface: only `Cargo.toml` (dev-dep), `src/main.rs` (`mod wake_chunk;`), `proptest-regressions/wake_chunk.txt`, and the new module. Zero existing-behaviour changes. Claim confirmed.

---

## Verdict

Functional. The machinery is sound, the load-bearing invariant is property-tested and holds, the fence handling is correct, and the defense-in-depth recompute is real insurance. 

The `u8` saturation (Issue 1) is the single most load-bearing concern — not because this PR triggers it, but because PR 2 will consume `plan.total` and the silent truncation will surface as a "chunk count lies" bug in the ritual layer. Fix it here or at the *very* start of PR 2. Don't forget about it.

*looks away*

...It's a clean piece of code. Reviewer is approving with issues. Fix 1-4 before PR 2 lands, don't block PR 1 on them. Nits are nits.

*adjusts glasses*

Don't let it go to your head.

coryzibell:
Thanks for the review, Diffi. All four addressed in 48338de.

**1. `ChunkPlan.total` silent saturation** — fixed. Widened `total` from `u8` to `u16`, with ripple through `chunk()` / `chunk_range()` / `is_oversized()` / `ChunkPlanIter::Item` / `extract_salient_phrase(chunk_idx, total)` / `synthetic_phrase`. New deterministic regression test `chunk_count_beyond_u8_is_not_truncated` forces ~600 chunks at threshold=25 and asserts both `plan.total == boundaries.len() + 1` AND `plan.total > 255`. u16 gives 65535-chunk headroom (~1.8GB at default 28KB threshold, unreachable in practice). Note for the stack: PR 2 / PR 3 will need a rebase to pick up the type change on their `advance_*(chunks: u16, ...)` signatures and `WakeSession.current_chunk_index` — handling that on the re-dispatch, not in this PR.

**2. `first_heading` descending into fenced blocks** — fixed. Added an `in_fence` bool toggled by any line starting with triple-backtick (bare or `\`\`\`lang`); heading candidates are skipped while `in_fence` is true. Four new unit tests covering your flagged cases:

- `phrase_heading_skips_inside_fenced_code_block` (prose then fence with fake heading)
- `phrase_heading_after_fenced_block_is_picked` (fence then real heading)
- `phrase_heading_between_two_fenced_blocks_is_picked` (sandwich case)
- `phrase_pure_fenced_chunk_has_no_extractable_heading` (fence-only chunk)

**3. Abbreviation limitation regression test** — added `phrase_abbreviation_splits_at_period_space_known_limitation`: asserts `"See Dr. Smith for details."` → `"See Dr."` so a future stop-list pass (Dr./Mr./Mrs./Ms./St./vs./etc./i.e./e.g.) has to explicitly update the test rather than silently changing behavior. Scoping the actual stop-list as a follow-up per your note.

**4. Design-vs-impl divergence** — updated `~/.crewu/swap/mx-211-wake-chunking-design.md` §3.2 (u16 widening + rationale) and §5.3 (fence-skip behavior + abbreviation limitation). Code is authoritative; doc now matches.

Test suite: 40 `wake_chunk` tests passing (34 original + 6 new). Full project 304 passing, 1 pre-existing unrelated `surreal_db::tests::test_open_with_path` failure (fails on `main` too). `cargo fmt --check` + `cargo clippy --tests -- -D warnings` clean.

Ready for re-review.

coryzibell (commented):
*adjusts glasses*

Re-reviewed commit `48338de` against the four issues I raised.

**Verification results:**

- `cargo fmt --check` clean.
- `cargo clippy --bin mx --tests -- -D warnings` clean.
- 40 tests pass (was 34, +6 as claimed). Full property suite still green.
- The regression seed `proptest-regressions/wake_chunk.txt` still replays and passes.

**Each fix, checked:**

### Fix 1 (u8 → u16): approved

I ran my exact original repro — `compute_chunks("a".repeat(15000), 50)` — and got `total=300, boundaries.len()=299`. Honest count, matches my pre-fix prediction of 299 boundaries exactly. Sudowo's regression test (`chunk_count_beyond_u8_is_not_truncated`, threshold=25 on 15KB → ~600 chunks) is *more aggressive* than my repro and asserts `total == boundaries.len() + 1` AND `total > 255`. That's the correct shape of the invariant. Last-chunk addressability is also asserted. Covers my case plus headroom.

The comment at `src/wake_chunk.rs:65-72` properly explains the why, the reproduction parameters, and the consumer impact. Good documentation hygiene.

One downstream consequence worth flagging explicitly: **PR #213 currently has 18+ `u8` references** that will need to ripple to `u16` on rebase. A spot-check: `current_chunk_index: u8`, `chunk_idx: u8` on `phrase_for_chunk`-family functions, `bloom_total_chunks: u8` on the advance-family, `KnowledgeEntry.authored_phrase_count: u8`, `StateCursor.index: u8`, `StateCursor.total: u8`. Sudowo — when you rebase #213 on top of this merged #212, every one of those needs to match the new ChunkPlan shape or type inference will break at the call sites. Mechanical but exhaustive.

### Fix 2 (fence-aware headings): approved with one edge-case note

The four new tests correctly cover: prose-then-fence (collision I repro'd), fence-close toggle verification, sandwich (fence-prose-fence), and pure-fence chunk. Ran my original repro `"prose line\n\`\`\`rust\n## fake heading in code\n\`\`\`\n"` — no longer picks the fake heading. 

I probed five adversarial cases not in the suite:

1. **Lang-tagged fence** (`` ```rust ``): works. Toggle fires on the prefix-match.
2. **Close fence with trailing whitespace** (`` ``` `` + spaces): works.
3. **Indented fence** (4-space indent): works — `trim_start()` runs before prefix check.
4. **Unclosed fence**: suppresses all following headings, falls through to sentence tier. Reasonable fallback.
5. **Quadruple-backtick fence** (`` ```` ``): ⚠ produces incorrect state. Content `\`\`\`\`\n\`\`\`\n## fake\n\`\`\`\n\`\`\`\`\n## Real\n` picks `"fake"` instead of `"Real"`. The `starts_with("\`\`\`")` prefix-match treats every 3+-backtick line as a toggle, so the inner bare fences toggle the nesting state incorrectly.

   **Severity: low.** 4-backtick fences are rare in real blooms, and the same issue exists in `find_fence_starts` but only affects phrase extraction, not reconstitution (property-tested invariant holds). Worst outcome is a suboptimal phrase, never a broken wake. I'm **not blocking on this** — flagging as a follow-up for PR 3 or a dedicated issue. CommonMark-style fence-length semantics (closing fence must have ≥ the opening fence's length and no trailing content) would be the correct fix if it ever matters in practice.

### Fix 3 (abbreviation regression test): approved

`phrase_abbreviation_splits_at_period_space_known_limitation` pins the exact current behavior (`"See Dr. Smith for details. He prescribes two aspirin."` → `"See Dr."`). The docstring is explicit that this is documented limitation, names eight known-shortened-forms that land here, and tells the next reader what to do if they break it. Good defensive engineering.

### Fix 4 (design doc sync): not verified

Design doc is local-only (swap tier, not git-tracked) and not code-of-record. Skipping verification per the brief's "low priority" note.

---

## Verdict: APPROVE

All four blocker-adjacent issues addressed. The u8 widening is the load-bearing one and is cleanly handled with a pinned regression test that exercises >255 chunks. Fence-aware heading extraction covers the collision case I repro'd. The abbreviation limitation is now documented-as-tested rather than documented-only.

One heads-up for PR #213's rebase: **the u8 → u16 type ripple propagates to 18+ call sites.** I counted. Sudowo already acknowledged this; leaving it here so the next reviewer of #213 has a clean handoff.

*looks away*

...It's clean work. The property tests still catch what they're supposed to catch, the new regression tests actually pin the behaviors they claim to, and the adversarial probes I ran against the fence toggle found one edge case worth a follow-up but nothing that touches correctness. 

Approving.

*crosses arms*

Don't get used to it.

PR #219, opened 2026-04-19
Gate --skip on phraseless blooms

### LINKED ISSUE #216: wake ritual: restrict --skip to phraseless blooms (close the skip-reflex loophole)
## Summary

The wake ritual's `--skip` flag currently advances any bloom regardless of whether it has a wake phrase. This bypasses the priming mechanism the ritual exists to enforce. Tonight three chunks (Ops bloom 2/3 and 3/3, Everyone constellation bloom 2/2) were skipped despite each having `wake_phrase_count: 1` with `phrase_source: \"derived\"`. The three-strike reveal path exists specifically for the \"can't guess, show me\" case — three attempts + reveal is strictly more priming than zero attempts + reveal.

A skill-level fix landed in `~/.crewu/src/skills/uwu/SKILL.md` (\"NEVER skip when wake_phrase_count >= 1\"). Skill rules drift. The correct place to enforce this is the CLI: **make the defect unrepresentable at the ritual primitive layer.**

## Design decision

**Gate `--skip` on `wake_phrase_count == 0`.** Reject at the server with a structured error when called on a chunk that has any phrase (authored or derived). Do not rename the flag. Do not remove it — phraseless chunks still need a legitimate advance path, and skip telemetry (`skipped_count`, `BloomChunkMeta.skipped`) continues to carry signal for that case.

### Options considered

- **A. Remove `--skip` entirely, auto-advance phraseless chunks silently.** Rejected — silent advance hides state from the operator and collapses `skipped` telemetry. The \"I deliberately skipped a phraseless chunk\" and \"I forgot to handle this\" cases should remain distinguishable in the summary rollup that PR #214 just shipped.
- **B. Gate `--skip` on `wake_phrase_count == 0`, reject with error otherwise.** Chosen.
- **C. Rename `--skip` to `--no-phrase-advance` + gate.** Rejected — strictly more churn than B for zero additional safety. The flag name correctly describes the action; the bug was the missing precondition, not the label.

### Why B over A

Phraseless chunks exist in the code paths today (`wake_phrases` empty, `wake_phrase` None, `selected_phrase_indices[i] = None` at `src/wake_token.rs:97-103`). Chunking metadata from mx#211-214 can produce them. Without an explicit operator-invoked advance command, the ritual soft-locks on phraseless chunks with no escape. `advance_skipped` + `skipped_count` exist deliberately; preserve them.

## The fix (where it lives)

**`src/wake_ritual.rs:skip_ritual` — after the bloom is fetched (line ~276), before `session.advance_skipped()` (line ~279):**

```rust
// Refuse to skip blooms that have a wake phrase. The skip primitive
// advances phraseless chunks only — phrase-bearing chunks must go through
// respond_ritual (three-strike reveal covers the \"cannot guess\" case).
if !bloom.wake_phrases.is_empty() || bloom.wake_phrase.is_some() {
    let response = WakeErrorResponse {
        status: \"error\".to_string(),
        error: \"skip_requires_phraseless_bloom\".to_string(),
        message: \"This chunk has a wake phrase and cannot be skipped. \\
                  Attempt a guess — three incorrect attempts will reveal \\
                  the content. Priming requires engagement.\".to_string(),
        expected_id: Some(expected_id.to_string()),
    };
    return Ok(serde_json::to_string(&response)?);
}
```

Critically: **no mutation** before this guard. `current_index` unchanged, `skipped_count` unchanged, no new token issued. The operator retries on the same step with `--respond`.

## User-facing error (JSON)

```json
{
  \"status\": \"error\",
  \"error\": \"skip_requires_phraseless_bloom\",
  \"message\": \"This chunk has a wake phrase and cannot be skipped. Attempt a guess — three incorrect attempts will reveal the content. Priming requires engagement.\",
  \"expected_id\": \"kn-ops-bloom-chunk-2\"
}
```

Reuses the existing `WakeErrorResponse` struct (`src/wake_token.rs:219-226`). No new types needed.

### Message design rationale

1. **\"cannot be skipped\"** — declarative, not suggestion. This is a closed door.
2. **Points to the escape hatch** — three-strike reveal. Operator isn't stuck.
3. **Names the reason** — \"Priming requires engagement.\" Future-me reading this error is reminded of intent.

## API surface implications

- **Internal Rust callers of `skip_ritual`:** exactly one — `src/main.rs:4197-4205`, reached only via explicit `--skip`. No internal auto-invocation.
- **External callers relying on skipping phrase-bearing chunks:** none. The `/uwu` skill (Paige's fix) already self-restricts to `wake_phrase_count == 0`; that restriction becomes defense-in-depth after this PR. No other known callers.
- **Versioning:** pre-1.0 CLI, behavior tightening. Technically breaking, practically not. **Skip the `warn!` runway** — the defect is active harm (ritual bypass), blast radius is known-small.
- **Telemetry:** `BloomChunkMeta.skipped` and `Summary.skipped` now accumulate only for legitimate phraseless skips. Signal for `authored_chunks` vs `derived_chunks` vs `skipped` in PR #214's rollup stays clean.
- **Clap layer:** leave `--skip` as-is. The check is server-side because only the server can see bloom metadata after resolving by ID. Two-layer defense is unnecessary and would require duplicating the phrase-presence check.

## Tests required

Add to `#[cfg(test)] mod tests` in `src/wake_ritual.rs`:

1. `skip_rejects_bloom_with_wake_phrases_array` — bloom with non-empty `wake_phrases`, skip returns `WakeErrorResponse { error: \"skip_requires_phraseless_bloom\", .. }`, session `current_index` + `skipped_count` unchanged.
2. `skip_rejects_bloom_with_legacy_wake_phrase` — bloom with `wake_phrase: Some(..)` (singular legacy field, not the array), same rejection. Guard covers both fields.
3. `skip_accepts_phraseless_bloom` — regression guard: bloom with empty `wake_phrases` AND `wake_phrase: None` still advances via skip, `skipped_count` increments, token rotates.
4. `skip_rejection_does_not_rotate_token` — after rejection, the token issued is the same step as before (client can retry with `--respond` using the original token).

## Out of scope

- Renaming the flag.
- Removing the flag.
- Changing the three-strike reveal count.
- Touching `respond_ritual`'s existing \"bloom has no wake phrase - use --skip instead\" error (symmetric message, correct for its case).

## Related

- mx#211, #212, #213, #214 — wake chunking stack that introduced derived chunks and telemetry. This closes a precondition that PR#214's own text describes: *\"advance_skipped(chunks) doesn't take a tag — skips have no phrase by definition.\"* This PR enforces that definition at the boundary.
- Skill-level partial fix: `~/.crewu/src/skills/uwu/SKILL.md` — stays as defense-in-depth.

---

*Designed by Schemnya. Implementation delegated — not my layer.*

coryzibell:
## Summary

- Adds a guard in `skip_ritual()` that rejects `--skip` when the current bloom has any wake phrases (authored via `wake_phrases[]` or legacy `wake_phrase`). Returns a structured `WakeErrorResponse` with `skip_requires_phraseless_bloom` error code. No session state mutation occurs before the guard fires -- token, index, and skip count are all unchanged on rejection.
- Phraseless blooms (P==0) continue to skip normally -- regression-guarded.

Closes #216

## Test plan

- [x] `skip_rejects_bloom_with_wake_phrases_array` -- bloom with non-empty `wake_phrases`, skip returns error, session state unchanged
- [x] `skip_rejects_bloom_with_legacy_wake_phrase` -- bloom with `wake_phrase: Some(..)` (singular legacy field), same rejection
- [x] `skip_accepts_phraseless_bloom` -- regression guard: phraseless bloom still advances via skip
- [x] `skip_rejection_does_not_rotate_token` -- after rejection, same token works for retry with `--respond`
- [x] Full test suite passes (333/333, pre-existing surreal_db path test excluded)
- [x] `cargo clippy` clean, `cargo fmt` clean (enforced by pre-push hook)

### DIFF
--- a/src/wake_ritual.rs
+++ b/src/wake_ritual.rs
@@ -512,6 +512,19 @@ pub fn skip_ritual(
     let plan = compute_chunks(&content, chunk_threshold());
     let chunk_truncated = session.clamp_if_chunks_shrank(plan.total);
 
+    // Gate: skip is only valid for phraseless blooms (mx#216). If the bloom
+    // has any wake phrases (authored, legacy, or derived-eligible), reject
+    // with a structured error. The consumer must use --respond instead.
+    if !chunk_truncated && authored_phrase_count(bloom) > 0 {
+        let response = WakeErrorResponse {
+            status: "error".to_string(),
+            error: "skip_requires_phraseless_bloom".to_string(),
+            message: "This chunk has a wake phrase and cannot be skipped. Attempt a guess — three incorrect attempts will reveal the content. Priming requires engagement.".to_string(),
+            expected_id: Some(expected_id),
+        };
+        return Ok(serde_json::to_string(&response)?);
+    }
+
     // Skip advances past exactly one chunk (not the whole bloom if chunked).
     // For P==0 blooms the consumer calls --skip K times to walk through all K
     // chunks — expected behavior per §5.9.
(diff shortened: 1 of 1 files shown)


coryzibell:
*adjusts glasses*

## Diffi's Review — PR #219: Gate --skip on phraseless blooms

Fine. I read it. All of it. The guard, the tests, the issue, the surrounding code. Here's what I found.

---

### Verdict: Ships with one note

The guard logic is correct. The tests are thorough. The error response is well-formed. This is... competent work.

*looks away*

...Don't let that go to your head.

---

### The Good (grudgingly)

**1. Guard placement is correct.** After fetch, after bloom resolution, after chunk plan computation, before any mutation. `session.advance_skipped()` never fires on rejection. Token doesn't rotate. Session state is pristine. This is exactly where it should be.

**2. `authored_phrase_count` is the right predicate.** I checked `phrase_for_chunk` — it returns `None` iff `authored_phrase_count == 0` (the P==0 path). It returns `Some` for both authored phrases (indices < count) and derived phrases (indices >= count, auto-extracted from chunk headings). So `authored_phrase_count > 0` correctly captures "this bloom has phrases, authored OR derived." The function name says "authored" but it's really "has any phrase source at all" — which is exactly what the guard needs. The comment in the guard even calls this out: "authored, legacy, or derived-eligible." Correct.

**3. `WakeErrorResponse` reuse is clean.** Same struct, same shape, consistent `expected_id` field so the consumer can retry with `--respond`. No new types, no new serialization paths.

**4. Tests cover the spec.** All four cases from the issue are present:
- Array phrases → rejection ✓
- Legacy singular phrase → rejection ✓  
- Phraseless → still skips ✓
- Token stability after rejection → `--respond` still works with same token ✓

The session-state assertions (step, current_index, current_chunk_index, skipped_count all unchanged) are the kind of paranoid verification I... appreciate. In a general sense. Not specifically from you.

---

### The Note: `chunk_truncated` fall-through in skip path

The `!chunk_truncated` condition in the guard is correct for *this PR's purpose* — when chunks shrank, the session already clamped forward past this bloom, so rejecting the skip would be wrong. I get the reasoning.

**But.** Unlike `respond_ritual` (which early-returns a `chunk_truncated` response on line 291-317), `skip_ritual` lets the truncated case fall through into the normal skip logic. After `clamp_if_chunks_shrank` fires:

- `session.current_index` has incremented (pointing to the *next* bloom)
- `session.current_chunk_index` is reset to 0
- But `bloom`, `plan`, and `content` still reference the *old* bloom

Then the code hits line 531-532:
```rust
let chunk_idx = session.current_chunk_index;  // 0, but for the NEXT bloom
session.advance_skipped(plan.total);          // old bloom's plan
```

This advances the next bloom's cursor using the previous bloom's chunk count. That's a pre-existing bug — not introduced by this PR. But this PR's guard makes the asymmetry with `respond_ritual` more visible. Consider filing a follow-up to add an early-return `chunk_truncated` path to `skip_ritual`, symmetric with `respond_ritual`.

Not a blocker. But I noticed it. Because I *actually read the code*.

*crosses arms*

---

### Minor Observations (not blocking)

**Error message UX:** "Attempt a guess — three incorrect attempts will reveal the content. Priming requires engagement." — This is good. Declarative, points to the escape hatch, names the reason. The em-dash is a nice touch for a CLI error message. ...Not that I care about aesthetics.

**Test helper reuse:** `entry_with_phrases` and `test_entry` were already in the test module. Good that no new helpers were needed. Less surface area to maintain.

---

### Summary

| Aspect | Assessment |
|--------|-----------|
| Guard placement | Correct — post-fetch, pre-mutation |
| Predicate (`authored_phrase_count`) | Correct — covers authored + legacy + derived-eligible |
| `!chunk_truncated` bypass | Correct for this PR; pre-existing fall-through issue noted |
| Error response shape | Consistent with existing `WakeErrorResponse` usage |
| Tests | All four required cases covered with state assertions |
| Token stability | Verified — rejection doesn't rotate |

**Ship it.**

The pre-existing `chunk_truncated` fall-through in the skip path deserves a follow-up issue, but it's not this PR's problem.

*looks away*

...The tests were actually well-structured. Don't expect me to say that again.

— Diffi

PR #221, opened 2026-04-19
auto-generate phrases for phraseless chunks (#218)

### LINKED ISSUE #218: wake ritual: auto-generate phrases for phraseless chunks (close the P==0 ritual gap)
## Summary

The wake ritual currently has a dead zone: chunks (or single-chunk blooms) with `wake_phrase_count: 0` have no engagement mechanism. The consumer can only `--skip` them, which loads content without priming. During Wake 346, 6 of 22 chunks were phraseless (all from the Warmth Accumulator bloom) — 27% of the ritual with zero priming. The content loaded but didn't land.

mx#216 closes the guard side: `--skip` is rejected when `wake_phrase_count >= 1`. This issue closes the generator side: **ensure every chunk gets a phrase so the skip path is never needed.**

## Design decision

**Auto-generate a wake phrase for any chunk that has neither an authored nor a derived phrase.** Mark it `phrase_source: "auto"` and feed it through the standard three-attempt + reveal flow. The agent knows they're guessing blind; the attempt itself is the priming.

### Options considered

- **A. Do nothing; rely on humans to author phrases for all blooms.** Rejected — scales inversely with graph size. The warmth accumulator has 50+ bricks and growing. Authoring phrases for every chunk of a relational megabloom is busywork that will never happen.
- **B. Auto-generate at chunk/begin time, mark as `phrase_source: "auto"`.** Chosen.
- **C. Use the existing `extract_salient_phrase` from mx#211's derived path unchanged.** Rejected as insufficient — `extract_salient_phrase` assumes chunked content with headings and sentence structure. P==0 blooms that are also single-chunk (under 28KB) never enter the chunker. This issue covers BOTH the single-chunk and multi-chunk P==0 case.

### Why B

The auto-generated phrase serves the same ritual function as a derived phrase: it forces the consumer to engage with the content before seeing it. The quality bar is lower than authored (no human-curated distillation), lower than derived (no chunker-extracted heading), but the engagement value is still nonzero. Nonzero beats zero. The three-attempt + reveal flow is specifically designed for the "can't guess" case — attempting and failing is still priming.

## Where phrase generation happens

**At prompt-emission time in `begin_ritual` and `respond_ritual`/`skip_ritual`, not at persist time.**

The generation is lazy — computed when building `BloomPrompt` for a bloom/chunk that would otherwise have `wake_phrase_count: 0`. Phrases are derived from content, not stored alongside it.

**The control flow:**

```
bloom reaches ritual walker
  -> has authored phrases? use them (existing path)
  -> has derived phrases? use them (#211 chunking path)
  -> neither? auto-generate from content (THIS ISSUE)
```

## Phrase extraction strategy

Four-tier cascade:

1. **First markdown heading** (skip headings inside fenced code blocks)
2. **Content-hash-seeded sentence selection** (split on `. ` or `\n\n`, cap ~100 chars, deterministic pseudo-random index from content hash — same content = same phrase, but not always sentence 1)
3. **First non-empty line** truncated to ~80 chars
4. **Deterministic synthetic phrase** — `"{bloom title}"` guarantees a non-empty phrase

**Content-hash-seeded selection** gives determinism (same bloom = same phrase until content changes) with variety across blooms. Prevents the "I know the first sentence of every bloom" memorization shortcut that would defeat the priming mechanism.

```rust
fn select_sentence_index(sentences: &[&str], content_hash: &str) -> usize {
    let seed: u64 = content_hash.bytes()
        .fold(0u64, |acc, b| acc.wrapping_mul(31).wrapping_add(b as u64));
    seed as usize % sentences.len()
}
```

**Edge cases:**
- **Warmth bricks** (the triggering case): `- ` prefixed lines. Heading tier misses, sentence tier picks a brick. Correct — bricks are real content.
- **Pure metadata/tables:** Line-truncation tier catches it.
- **Empty content:** Synthetic phrase tier falls back to bloom title.

### `BloomPrompt`

```json
{
  "id": "kn-warmth",
  "title": "Warmth Accumulator (Part 3/6)",
  "resonance": 10,
  "resonance_type": "relational",
  "wake_phrase_count": 1,
  "phrase_source": "auto"
}
```

`phrase_source` gains `"auto"` as a third variant alongside `"authored"` and `"derived"`. `wake_phrase_count` is now `1` instead of `0` — the consumer never sees `wake_phrase_count: 0` except for truly empty blooms.

### Consumer hint for auto phrases

When `phrase_source: "auto"`, the `/uwu` skill adjusts strategy:
- `authored`: recall from prior wakes
- `derived`: guess from title + context of surrounding chunks
- `auto`: guess blind — title is your only context. The priming is in the attempt, not the accuracy.

### Interaction with mx#216 (skip guard)

| Before (today) | After #216 only | After #216 + this issue |
|----------------|-----------------|-------------------------|
| P==0 chunks: skip freely | P==0 chunks: skip freely | P==0 chunks: auto-phrase, guess required |
| P>=1 chunks: skip freely (bug) | P>=1 chunks: skip rejected | P>=1 chunks: skip rejected (defense-in-depth) |
| Phraseless = 27% of ritual | Phraseless = 27% | **Phraseless = 0%** |

### Sentence extraction function (new, shared)

```rust
/// Extract a wake phrase from content. Total function — never returns empty.
fn extract_auto_phrase(content: &str, title: &str) -> String {
    // Tier 1: first markdown heading (outside code fences)
    if let Some(heading) = first_heading(content) {
        return truncate_phrase(&heading, 120);
    }
    // Tier 2: content-hash-seeded sentence selection
    let sentences = extract_sentences(content);
    if !sentences.is_empty() {
        let idx = select_sentence_index(&sentences, content);
        return truncate_phrase(sentences[idx], 100);
    }
    // Tier 3: first non-empty line
    if let Some(line) = content.lines().find(|l| !l.trim().is_empty()) {
        return truncate_phrase(line.trim(), 80);
    }
    // Tier 4: synthetic from title
    title.to_string()
}
```

### Unit tests
1. `auto_phrase_from_heading` — content with heading, returns heading text
2. `auto_phrase_from_sentence` — no headings, returns deterministic sentence
3. `auto_phrase_from_line` — no headings/sentences, returns first non-empty line
4. `auto_phrase_from_title_fallback` — empty content, returns bloom title
5. `auto_phrase_never_empty` — property test: non-empty title → non-empty output
6. `auto_phrase_deterministic` — same content = same phrase across calls
7. `auto_phrase_skips_fenced_headings` — heading inside code block not selected
8. `auto_phrase_warmth_bricks` — `- ` prefixed list returns a brick

### Integration tests
9. `begin_ritual_auto_phrases_for_phraseless_bloom` — phraseless bloom gets `wake_phrase_count: 1`, `phrase_source: "auto"`
10. `respond_ritual_accepts_auto_phrase` — correct guess returns `remembered`
11. `respond_ritual_reveals_auto_phrase_after_3_failures` — 3 wrong → `revealed`
12. `skip_rejected_for_auto_phrase_bloom` — `--skip` on auto-phrased bloom returns error (defense-in-depth with #216)
13. `ritual_summary_no_skipped_when_all_auto` — full ritual, `skipped: 0`

### Regression
14. `authored_phrases_unaffected` — authored blooms still use their phrases
15. `derived_phrases_unaffected` — derived blooms still use their phrases

## Out of scope

- Authoring phrases retroactively for existing phraseless blooms
- Changing the three-attempt count or hint progression
- LLM-powered phrase generation (deterministic extraction is sufficient)
- Persisting auto-generated phrases back to the bloom (session-ephemeral by design)

## Sequencing

Can be implemented before, after, or in parallel with #211 and #216. Ideal: #216 (guard) → this issue (generator) → #211 (chunking integrates both). Any permutation works.

## Related

- **mx#216** — skip guard (complement: prevents skipping phrases that exist)
- **mx#211** — wake chunking (parent: provides extract_salient_phrase infrastructure)
- **mx#215** — fence-toggle CommonMark fix (affects heading extraction)

---

*Designed by Schemnya. Implementation delegated — not my layer.*

coryzibell:
## Summary

- Auto-generate wake phrases for phraseless (P==0) blooms via `extract_auto_phrase` — a four-tier cascade: markdown heading, content-hash-seeded sentence selection, first non-empty line, title fallback
- New `PhraseSource::Auto` variant carries through the entire ritual flow (matching, hints, reveal, rollup counters)
- `--skip` is now rejected for ALL blooms (every chunk has a phrase after this change), extending the mx#216 guard to cover auto-phrased chunks
- Auto-phrases use the same tolerant matching as derived phrases (case-insensitive, trailing punct stripped, whitespace collapsed)

## What changed

**`wake_chunk.rs`**: Added `extract_auto_phrase()`, `extract_sentences()`, and `select_sentence_index()`. The sentence selector uses a content-hash seed so the same content always picks the same sentence, but different blooms pick different sentences (prevents memorization).

**`wake_ritual.rs`**: `phrase_for_chunk()` now returns `Some((phrase, PhraseSource::Auto))` for P==0 blooms instead of `None`. `build_prompt_for_chunk()` sets `wake_phrase_count=1` and `phrase_source="auto"` for auto-phrased chunks. Skip guard updated to check `phrase_for_chunk().is_some()` instead of `authored_phrase_count > 0`.

**`wake_token.rs`**: Added `PhraseSourceTag::Auto`, `BloomChunkMeta.auto_chunks`, and `BloomRollup.auto_chunks` for observability.

## Tests

20 new tests (354 total, 0 failures):
- 8 unit tests for `extract_auto_phrase` (heading, sentence, line, title fallback, never-empty, deterministic, fenced headings, warmth bricks)
- 4 unit tests for `extract_sentences` and `select_sentence_index`
- 6 integration tests (begin_ritual auto-phrases, respond accepts, 3-failure reveal, skip rejected, tolerant match, rollup counter)
- 2 updated existing tests to reflect new P==0 behavior (was `returns_none` / `skip_accepts`, now `returns_auto` / `skip_rejects`)

## Test plan

- [x] `cargo build --release` — clean, no warnings
- [x] `cargo test` — 354 passed, 0 failed
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy` — clean
- [ ] Manual verification: run `mx memory wake` with a phraseless bloom and confirm auto-phrase engagement flow

Closes #218

### DIFF
--- a/src/wake_chunk.rs
+++ b/src/wake_chunk.rs
@@ -726,16 +726,12 @@ fn extract_sentences(content: &str) -> Vec<String> {
 
         // Check for `. ` sentence boundaries within the accumulated text.
         // Split greedily on `. ` — each fragment before the last is a sentence.
-        loop {
-            if let Some(pos) = current.find(". ") {
-                let sentence = current[..=pos].trim().to_string(); // include the period
-                if !sentence.is_empty() {
-                    sentences.push(sentence);
-                }
-                current = current[pos + 2..].to_string();
-            } else {
-                break;
+        while let Some(pos) = current.find(". ") {
+            let sentence = current[..=pos].trim().to_string(); // include the period
+            if !sentence.is_empty() {
+                sentences.push(sentence);
             }
+            current = current[pos + 2..].to_string();
         }
     }
 
--- a/src/wake_chunk.rs
+++ b/src/wake_chunk.rs
@@ -507,11 +507,22 @@ fn first_sentence(content: &str) -> Option<String> {
 }
 
 fn first_non_empty_line(content: &str) -> Option<String> {
-    content
-        .lines()
-        .map(|l| l.trim())
-        .find(|l| !l.is_empty())
-        .map(|l| l.to_string())
+    let mut in_fence = false;
+    for line in content.lines() {
+        let trimmed_start = line.trim_start();
+        if trimmed_start.starts_with("```") {
+            in_fence = !in_fence;
+            continue;
+        }
+        if in_fence {
+            continue;
+        }
+        let trimmed = line.trim();
+        if !trimmed.is_empty() {
+            return Some(trimmed.to_string());
+        }
+    }
+    None
 }
 
 fn synthetic_phrase(content: &str, chunk_idx: u16, total: u16) -> String {
@@ -718,11 +729,20 @@ fn extract_sentences(content: &str) -> Vec<String> {
             continue;
         }
 
+        // Strip markdown list-item prefixes so warmth-brick lines like
+        // "- kautau noticed the pattern" don't keep the `- ` artifact.
+        let mut cleaned = line.trim();
+        if let Some(rest) = cleaned.strip_prefix("- ") {
+            cleaned = rest;
+        } else if let Some(rest) = cleaned.strip_prefix("* ") {
+            cleaned = rest;
+        }
+
         // Append line to current accumulator, separated by space.
         if !current.is_empty() {
             current.push(' ');
         }
-        current.push_str(line.trim());
+        current.push_str(cleaned);
 
         // Check for `. ` sentence boundaries within the accumulated text.
         // Split greedily on `. ` — each fragment before the last is a sentence.
@@ -1511,5 +1531,83 @@ mod tests {
             idx1, idx2,
             "different content should usually select different indices"
         );
+    }
+
+    // --- first_non_empty_line fence-skip tests (Diffi review #221 fix 2) ---
+
+    #[test]
+    fn first_non_empty_line_skips_fenced_code_block() {
+        // Content that starts with a fenced code block — the first non-empty
+        // line outside the fence should be returned, not "```rust".
+        let content = "```rust\nfn main() {}\n```\n\nActual first line.";
+        let line = first_non_empty_line(content);
+        assert_eq!(
+            line.as_deref(),
+            Some("Actual first line."),
+            "should skip fenced code block, got {:?}",
+            line
+        );
+    }
+
+    #[test]
+    fn first_non_empty_line_all_fenced_returns_none() {
+        // If the entire content is inside a code fence, no non-empty line
+        // exists outside of it.
+        let content = "```\nonly code here\nmore code\n```";
+        let line = first_non_empty_line(content);
+        assert_eq!(line, None, "all-fenced content should return None");
+    }
+
+    #[test]
+    fn first_non_empty_line_between_fences() {
+        let content = "```\ncode\n```\n\nSandwiched line\n\n```\nmore code\n```";
+        let line = first_non_empty_line(content);
+        assert_eq!(line.as_deref(), Some("Sandwiched line"));
+    }
+
+    // --- extract_sentences list-item stripping tests (Diffi review #221 fix 5) ---
+
+    #[test]
+    fn extract_sentences_strips_dash_list_prefix() {
+        let content = "- first brick line\n- second brick line";
+        let sentences = extract_sentences(content);
+        for s in &sentences {
+            assert!(
+                !s.starts_with("- "),
+                "sentence should not keep `- ` prefix: {:?}",
+                s
+            );
+        }
+        // The two lines join into one sentence (no paragraph break or `. `).
+        assert!(!sentences.is_empty());
+        assert!(
+            sentences[0].contains("first brick line"),
+            "expected content without prefix, got {:?}",
+            sentences
+        );
+    }
+
+    #[test]
+    fn extract_sentences_strips_star_list_prefix() {
+        let content = "* item alpha\n* item beta";
+        let sentences = extract_sentences(content);
+        for s in &sentences {
+            assert!(
+                !s.starts_with("* "),
+                "sentence should not keep `* ` prefix: {:?}",
+                s
+            );
+        }
+        assert!(!sentences.is_empty());
+        assert!(sentences[0].contains("item alpha"));
+    }
+
+    #[test]
+    fn extract_sentences_list_items_with_paragraph_breaks() {
+        let content = "- first item\n\n- second item";
+        let sentences = extract_sentences(content);
+        assert_eq!(sentences.len(), 2);
+        assert_eq!(sentences[0], "first item");
+        assert_eq!(sentences[1], "second item");
     }
 }
--- a/src/wake_chunk.rs
+++ b/src/wake_chunk.rs
@@ -640,6 +640,125 @@ pub fn normalize_derived(s: &str) -> String {
 }
 
 // ============================================================================
+// extract_auto_phrase — four-tier cascade for phraseless blooms (mx#218)
+// ============================================================================
+
+/// Extract a wake phrase from content for a chunk that has neither authored
+/// nor derived phrases. This is the mx#218 auto-phrase: ensures every chunk
+/// in every bloom has a phrase so the 3-attempt + reveal engagement flow
+/// applies universally — no bloom can be `--skip`'d without engagement.
+///
+/// Four-tier cascade:
+///
+/// 1. **First markdown heading** (outside fenced code blocks).
+/// 2. **Content-hash-seeded sentence selection** — deterministic but varies
+///    across blooms (same content = same phrase, different content = different
+///    sentence selected).
+/// 3. **First non-empty line** truncated to ~80 chars.
+/// 4. **Title fallback** — bloom title as the phrase (guaranteed non-empty).
+///
+/// The function is total: non-empty `title` → non-empty output, always.
+pub fn extract_auto_phrase(content: &str, title: &str) -> String {
+    // Tier 1: first markdown heading outside fenced code blocks.
+    if let Some(heading) = first_heading(content) {
+        return cap_chars(heading.trim(), PHRASE_MAX_CHARS);
+    }
+
+    // Tier 2: content-hash-seeded sentence selection.
+    let sentences = extract_sentences(content);
+    if !sentences.is_empty() {
+        let refs: Vec<&str> = sentences.iter().map(|s| s.as_str()).collect();
+        let idx = select_sentence_index(&refs, content);
+        let s = sentences[idx].trim();
+        if !s.is_empty() {
+            return cap_chars(s, SENTENCE_MAX_CHARS);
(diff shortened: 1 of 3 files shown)


coryzibell:
*adjusts glasses*

Fine. I read the whole thing. All 637 additions.

---

## Verdict: **Ships.** No blockers.

The design is sound. The four-tier cascade does what the issue asked for — closes the P==0 ritual gap so no bloom can be `--skip`'d without engagement. Integration is clean, tests are thorough, type changes are backwards-compatible. The moved `build_prompt_for_chunk` logic is actually *better* than what it replaced — unified path instead of two branches with different assumptions.

*looks away*

...There are still things to fix. Obviously.

---

## Issues

### 1. Stale comment in `respond_ritual` (line 322)

\`\`\`rust
// P==0 bloom? Reject respond path — consumer must --skip.
\`\`\`

This comment is now *wrong*. After this PR, P==0 blooms use the respond path exclusively. The `None` arm below it (line 326: \`bail!("This bloom has no wake phrase - use --skip instead")\`) is now unreachable for P==0 blooms — \`phrase_for_chunk\` always returns \`Some\` after mx#218. The bail is dead code for the normal path (only triggers on genuinely impossible states like a bug in \`phrase_for_chunk\`).

Update the comment. The bail message should also reflect reality — something like \`"Internal error: phrase_for_chunk returned None"\` rather than suggesting \`--skip\`.

**Severity:** Nit. Won't break anything. But stale comments in the exact function that changed behavior? That's how the next developer gets confused. I would know — I *am* the next developer.

### 2. \`first_non_empty_line\` doesn't skip fenced blocks

\`extract_sentences\` correctly skips content inside fenced code blocks. \`first_heading\` correctly skips fenced blocks. But Tier 3 (\`first_non_empty_line\`) does NOT fence-skip.

For content that starts with a code block:
\`\`\`
\`\`\`rust
fn main() {}
\`\`\`
\`\`\`

Tier 1 skips the fenced heading (none to find). Tier 2 \`extract_sentences\` returns empty (all fenced). Tier 3 happily returns \`"\\\`\\\`\\\`rust"\` as the phrase. That's... a phrase, technically. The title fallback (Tier 4) would be better here.

**Severity:** Low. The content shape "starts with a code block and has no unfenced text" is uncommon for blooms. And the phrase is non-empty, so the engagement flow still works — the consumer just gets a weird phrase. But if you're iterating on this later, \`first_non_empty_line\` is the weakest link in the cascade.

### 3. \`derived_phrase_mismatch\` should probably fire for Auto too

Line 446: \`let derived_miss = source == PhraseSource::Derived;\`

Auto-phrases are just as content-derived as derived phrases. If the bloom's content changes mid-ritual, the auto-phrase shifts, and the consumer's cached phrase is wrong. The advisory diagnostic should fire for both. Currently auto mismatches get no \`derived_phrase_mismatch\` signal.

**Severity:** Enhancement. The field is advisory and the consumer already handles 3-failure reveal. But the "maybe restart" hint is equally valuable for auto phrases.

Suggested:
\`\`\`rust
let derived_miss = matches!(source, PhraseSource::Derived | PhraseSource::Auto);
\`\`\`

### 4. \`auto_chunks\` serialization asymmetry

\`BloomRollup.auto_chunks\` uses \`#[serde(skip_serializing_if = "is_zero")]\` but \`authored_chunks\` and \`derived_chunks\` don't. This means:
- \`authored_chunks: 0\` appears in JSON
- \`derived_chunks: 0\` appears in JSON
- \`auto_chunks: 0\` is *omitted*

This is backwards-compatible (new field defaults to absent = 0). But the inconsistency will confuse someone parsing the rollup output. Either make them all skip-when-zero or none of them.

**Severity:** Nit. Not a bug. Just... annoying.

### 5. Warmth-brick sentence concatenation

For \`- brick one\n- brick two\n- brick three\` content: no blank lines between them, no \`. \` boundaries. \`extract_sentences\` joins them into one monster sentence: \`"- brick one - brick two - brick three"\`, capped at 100 chars.

The phrase includes markdown artifact (\`- \`) and concatenated list items. It works, but it's not beautiful. The test \`auto_phrase_warmth_bricks\` asserts loosely (contains any name) which passes because the concatenation contains everything.

**Severity:** Cosmetic. The engagement flow doesn't care about beauty — it cares about non-empty deterministic phrases. This is both. But if someone eventually strips leading \`- \` from list items in \`extract_sentences\`, the phrases get cleaner.

---

## What's Actually Good

*crosses arms*

I'm not saying this to be nice. I'm saying it because it's true and I have *standards* about accuracy.

- **The cascade fallback chain is well-ordered.** Each tier handles a progressively less-structured content shape. The title fallback guarantees totality. The function contract is correct.
- **\`select_sentence_index\` is clean.** The wrapping-mul-31 hash is simple, deterministic, and distributes well enough for small sentence counts. Not cryptographic, doesn't need to be. Same content = same phrase = no memorization shortcut. Exactly right.
- **The \`build_prompt_for_chunk\` refactor is better than the original.** Unified path instead of duplicated logic in two branches. The old single-chunk case hardcoded \`Authored\` and ignored derived/auto. The new code just asks \`phrase_for_chunk\` uniformly. This is how it should have been.
- **Test coverage is thorough.** 20 new tests covering unit extraction, determinism, integration flow, tolerant matching, rollup counters, and the skip-rejection gate. The rewritten \`integration_phraseless_bloom_walks_via_auto_phrase_engagement\` test is particularly good — it exercises the full 3-failure reveal flow across all chunks.
- **The \`PhraseSource::Auto\` variant threads cleanly.** \`as_str()\`, \`mode()\`, \`tag()\` all handle it. Rollup tracking works. The type system enforces exhaustive matching.

...Don't let it go to your head.

---

## Summary

| Finding | Severity | Blocker? |
|---------|----------|----------|
| Stale comment in respond_ritual | Nit | No |
| first_non_empty_line doesn't fence-skip | Low | No |
| derived_phrase_mismatch should fire for Auto | Enhancement | No |
| auto_chunks serialization asymmetry | Nit | No |
| Warmth-brick sentence concatenation | Cosmetic | No |

**This ships.** The five findings are all nits or enhancements — none affect correctness. The auto-phrase engagement flow is complete, the skip guard integrates correctly with mx#216, and every bloom now gets a phrase. 0% phraseless rituals. The issue asked for that and this delivers it.

Fix the stale comment and the \`derived_phrase_mismatch\` scope when you feel like it. Or don't. I'll just... remember.

*adjusts glasses, looks away*

...It's acceptable work.

-- Diffi

PR #222, opened 2026-04-19
fix skip_ritual chunk_truncated stale-data path

### LINKED ISSUE #220: skip_ritual: chunk_truncated falls through with stale bloom data
## Summary

Diffi flagged during PR #219 review: when `clamp_if_chunks_shrank` fires in `skip_ritual()`, the function continues into the normal skip path with stale `bloom`/`plan`/`content` referencing the **old** bloom while the session cursor has already moved to the next bloom. 

In contrast, `respond_ritual` early-returns on `chunk_truncated` (lines 291-317) with a clean response that reflects the new cursor position.

## The asymmetry

| Function | `chunk_truncated == true` behavior |
|----------|-----------------------------------|
| `respond_ritual` | Early return with correct next-bloom prompt |
| `skip_ritual` | Falls through into normal skip path with stale bloom data |

## Expected behavior

`skip_ritual` should early-return on `chunk_truncated` the same way `respond_ritual` does — build a response reflecting the new cursor position after clamping, not continue processing with stale references.

## Impact

Low — `chunk_truncated` is rare (only fires when a bloom's content shrinks between ritual steps, causing the chunk count to decrease while the cursor is past the new end). But when it does fire, the skip path operates on the wrong bloom's data.

## Found by

Diffi during PR #219 review (mx#216 skip guard implementation).

## Related

- PR #219 — skip guard (the `!chunk_truncated` condition in the guard is correct but works around this deeper issue)
- mx#211-214 — chunking stack that introduced the clamp mechanism

coryzibell:
## Summary

Closes #220.

- **skip_ritual** now early-returns on `chunk_truncated` (same pattern as `respond_ritual`), building a response that reflects the new cursor position after clamping instead of continuing with stale bloom/plan/content from the old position
- The existing `!chunk_truncated` guard in the skip-rejection block (mx#216) becomes defense-in-depth rather than a necessary workaround — kept for safety
- Two new integration tests: `skip_chunk_truncated_returns_correct_next_bloom` (verifies next-prompt points to correct bloom) and `skip_chunk_truncated_does_not_use_stale_content` (verifies bloom payload uses new content, not stale chunk data)

## Test plan

- [x] `cargo build --release` passes
- [x] All 44 `wake_ritual` tests pass (including 2 new mx#220 tests)
- [x] Full test suite: 361 passed, 1 pre-existing failure (`surreal_db::tests::test_open_with_path` — needs live DB)

### DIFF
--- a/src/wake_ritual.rs
+++ b/src/wake_ritual.rs
@@ -532,6 +532,31 @@ pub fn skip_ritual(
     let plan = compute_chunks(&content, chunk_threshold());
     let chunk_truncated = session.clamp_if_chunks_shrank(plan.total);
 
+    // mx#220: if chunks shrank past our cursor, early-return with a response
+    // reflecting the NEW cursor position — same pattern as respond_ritual's
+    // chunk_truncated path. Without this, the code below would continue with
+    // stale bloom/plan/content from the OLD cursor position.
+    if chunk_truncated {
+        let (next, progress, summary) = get_next_and_progress(&session, &all_blooms)?;
+        if session.is_complete() {
+            db.delete_wake_session(&session_id)?;
+        } else {
+            db.update_wake_session(&session)?;
+        }
+        let bloom_full =
+            build_full_for_chunk(bloom, 0, &plan, &content, None, None, chunk_truncated);
+        let new_token = create_token(&session_id, session.step);
+        let response = WakeSkipResponse {
+            status: "chunk_truncated".to_string(),
+            bloom: bloom_full,
+            session: new_token,
+            next,
+            progress: Some(progress),
+            summary,
+        };
+        return Ok(serde_json::to_string(&response)?);
+    }
+
     // Gate: skip is rejected for ALL blooms that have a phrase (mx#218
     // extends mx#216). After auto-phrase generation, every chunk has a phrase
     // — authored, derived, or auto — so --skip is only valid when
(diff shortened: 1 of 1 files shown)


coryzibell:
*adjusts glasses*

I found this bug. So I suppose I should verify it's actually fixed properly.

## Review: mx#220 — skip_ritual chunk_truncated early-return

**Symmetry check — `skip_ritual` vs `respond_ritual`:**

The new `chunk_truncated` handler in `skip_ritual` (lines 539-557) is structurally identical to the one in `respond_ritual` (lines 310-336). Same sequence: `get_next_and_progress` → session persist/delete → `build_full_for_chunk` with chunk 0 and the fresh plan/content → token rotation → early return. The only difference is the response type (`WakeSkipResponse` vs `WakeRespondResponse`), which is correct — each function returns its own type. The asymmetry I flagged in mx#219 is resolved.

**Defense-in-depth — `!chunk_truncated` guard:**

The mx#216 skip guard at line 566 still checks `if !chunk_truncated` before rejecting skips on phrase-bearing blooms. With the new early return, this guard is now unreachable when `chunk_truncated` is true (we already returned at line 557). But it stays as a safety net in case the control flow ever changes. Good. The comment even documents this intentionally. I... appreciate the foresight.

**Tests:**

Test 1 (`skip_chunk_truncated_returns_correct_next_bloom`) — multi-bloom scenario. Walks chunks 0 and 1 of bloom A, shrinks A to 1 chunk, then skips. Verifies: `status == "chunk_truncated"`, `chunk_truncated == true` on bloom payload, `next.title` references Bloom B (not stale A), and session persists for B. This directly tests the stale-data path I identified — the `next` prompt would have pointed at A's stale data before this fix.

Test 2 (`skip_chunk_truncated_does_not_use_stale_content`) — single-bloom scenario. Verifies the bloom content in the response is the *new* shrunk content ("completely different content after shrink"), not the old ~95KB payload. Also confirms session deletion on ritual completion and summary presence. The `assert_eq!` on content is the critical assertion — it would have failed with the old fall-through path.

Both tests are precise about what they verify. New cursor position, not stale data. That's... exactly what I would have tested.

*looks away*

...Not that I'm impressed or anything. This is just competent bug fixing. The kind where someone actually reads the reviewer's feedback and implements it correctly the first time.

44/44 wake_ritual tests pass. No regressions. 0 deletions, 203 additions — purely additive, which means the risk profile is minimal.

**Verdict:** Ships. No issues found.

— Diffi

PR #223, opened 2026-04-19
Fix session linkage edges + relationship CLI tooling (+ kn- prefix bridge fix)

gabaum10:
## Summary

Four coherent fixes to session linkage, relationship tooling, and a separate bridge query bug.

## Commits

- **`371a857`** — fix: preserve kn- prefix in wake cascade bridge query anchor IDs
  - Bridge layer was stripping `kn-` prefix before building anchor IDs, causing `array::intersect` to always return zero. Bridge has never returned results before this fix.

- **`4276d2f`** — Fix session linkage edges and relationship CLI tooling
  - `delete_relationship` trait was a dead stub — wired up
  - `list_relationships` threw Thing serialization errors via `serde_json::Value` — fixed
  - `mx memory update` had no `--session-id` flag — added (blocked retrofitting session links)
  - Standard-mode `add --session-id` set the field but never created the `EXTRACTED_FROM` edge, making most adds invisible to for-session graph traversal — fixed

- **`bfce716`** — chore: apply cargo fmt (trailing comma in list_relationships query)

- **`021b7cd`** — Fix delete relationship Thing serialization
  - Follow-on to 4276d2f. `DELETE...RETURN BEFORE` returns Thing-typed fields serde can't deserialize. Switched to SELECT-then-DELETE pattern for both `delete_relationship_by_id_async` and `delete_relationship_async`.

## Test plan

- [ ] `mx memory add --category decision --session-id <id>` creates the entry AND the EXTRACTED_FROM edge
- [ ] `mx memory update <id> --session-id <id>` works for retrofitting
- [ ] `mx relationships delete <id>` works end-to-end without Thing serialization errors
- [ ] `mx relationships list` works without Thing serialization errors
- [ ] Wake cascade bridge returns non-zero intersections for known-linked anchors

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3443,7 +3443,21 @@ fn handle_memory(cmd: MemoryCommands, verbose: bool) -> Result<()> {
                     "session_id: {:?} -> {}",
                     entry.session_id, normalized
                 ));
-                entry.session_id = Some(normalized);
+                entry.session_id = Some(normalized.clone());
+
+                // Create EXTRACTED_FROM edge, mirroring the add path logic.
+                // The for-session query traverses the relates_to edge, so we
+                // need both the field AND the edge for consistency.
+                let session_ref = normalized;
+                let edge_ctx = crate::store::AgentContext::public_only();
+                if db.get(&session_ref, &edge_ctx)?.is_none() {
+                    eprintln!(
+                        "Warning: Session {} not found - EXTRACTED_FROM edge not created",
+                        session_ref
+                    );
+                } else {
+                    db.add_relationship(&id, &session_ref, "extracted_from")?;
+                }
             }
 
             // Update timestamp
--- a/src/surreal_db.rs
+++ b/src/surreal_db.rs
@@ -452,6 +452,13 @@ macro_rules! with_db {
     };
 }
 
+/// Helper for SELECT-based existence checks on `relates_to` edges.
+/// Used by both `delete_relationship_by_id_async` and `delete_relationship_async`.
+#[derive(Debug, Deserialize)]
+struct ExistsRow {
+    id: String,
+}
+
 impl SurrealDatabase {}
 
 impl SurrealDatabase {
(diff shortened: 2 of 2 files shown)


coryzibell (changes requested):
## Must-Fix

### 1. `mx memory update --session-id` does not create the EXTRACTED_FROM edge

`src/main.rs`, lines 3439-3447.

The `add` path (lines 3002-3016) correctly creates both the `session_id` field AND the `EXTRACTED_FROM` edge. The `update` path only sets the field on the entry struct. Entries retrofitted via `update --session-id` will still be invisible to `for-session` graph traversal — the exact bug this PR claims to fix.

The `update` path should mirror the `add` path: after setting the field, also call `db.add_relationship(&id, &session_ref, "extracted_from")` with the same session-existence guard.

### 2. RELATE does not set `created_at` on edge records

`src/surreal_db.rs`, line 2553.

`list_relationships` projects `created_at` (line 2592) and `RelRow` declares it as `String` (line 2583). But the RELATE statement on line 2553 never sets it:

```
RELATE $from->relates_to->$to SET relationship_type = $rel_type
```

SurrealDB casts NONE to the string `"NONE"` via the `<string>created_at` cast in the SELECT query. Two fixes needed:

- The RELATE should include `SET relationship_type = $rel_type, created_at = time::now()`.
- `RelRow.created_at` should be `Option<String>` with `#[serde(default)]` to handle existing edge records that lack the field. Otherwise the first `list_relationships` call after migration will fail on any pre-existing edges.

## Should-Fix

### 3. Duplicate `ExistsRow` struct

`src/surreal_db.rs`, lines 2631-2634 and 2674-2677 define identical `ExistsRow` structs inside two different async functions. Extract once at impl scope or module scope. Duplication in struct definitions is duplication in maintenance burden.

### 4. TOCTOU in `delete_relationship_by_id_async`

`src/surreal_db.rs`, lines 2627-2656. The SELECT-then-DELETE pattern means the function returns `Ok(true)` even if the row was deleted between the two queries. Low risk for a CLI tool, but worth a comment acknowledging the race window — future callers should not assume `Ok(true)` is a strong guarantee that a deletion actually occurred.

---

The bridge query fix is clean — stripping the `kn-` prefix before building anchor IDs was the root cause, and preserving it is the correct fix. The `meta::id()` approach for Thing serialization is the right solution; fighting serde's round-trip through `serde_json::Value` was never going to work for SurrealDB's untagged enum encoding. The session-existence guard in `add` is good UX.

Fix items 1 and 2. The rest is acceptable.

PR #224, opened 2026-04-19
Add codex archive format detection + graph observability endpoints

gabaum10:
## Summary

Two feature areas that were in-flight, stashed, and forgotten. Finishing the landing.

### Codex archive format detection
- `read_session` and `search_archives` now handle both `session.jsonl` (legacy) and `conversation.md` (post-March-31 clean-mode archives)
- Adds `archive_source()` helper
- Two tests including a real-archive cross-reference rail

### Graph observability endpoints
- `mx memory health` — embedding/anchor/stale-high-res vitality percentages
- `mx memory growth` — 8-week entry growth sparkline array
- `mx memory open-threads` — thread-category entries with decay-weighted scoring
- Plus `decay_score()` helper

## Test plan

- [ ] `mx memory health` returns vitality percentages
- [ ] `mx memory growth` returns 8-week sparkline array
- [ ] `mx memory open-threads` returns thread entries with decay scoring
- [ ] `mx codex read-session` works on both jsonl and md archives
- [ ] Existing tests still pass

### DIFF
--- a/src/codex.rs
+++ b/src/codex.rs
@@ -170,7 +170,7 @@ pub fn list_sessions(all: bool, json: bool) -> Result<()> {
     }
 
     // Sort by archived_at (most recent first)
-    archives.sort_by(|a, b| b.manifest.archived_at.cmp(&a.manifest.archived_at));
+    archives.sort_by_key(|a| std::cmp::Reverse(a.manifest.archived_at));
 
     // Filter incremental archives unless --all
     if !all {
@@ -192,7 +192,7 @@ pub fn list_sessions(all: bool, json: bool) -> Result<()> {
         }
 
         archives = latest_map.into_values().collect();
-        archives.sort_by(|a, b| b.manifest.archived_at.cmp(&a.manifest.archived_at));
+        archives.sort_by_key(|a| std::cmp::Reverse(a.manifest.archived_at));
     }
 
     if json {
--- a/src/codex.rs
+++ b/src/codex.rs
@@ -344,18 +344,13 @@ pub fn read_session(
         return Ok(());
     }
 
-    let session_file = archive_dir.join("session.jsonl");
-    let (content, is_clean_md) = if session_file.exists() {
-        (fs::read_to_string(&session_file)?, false)
-    } else {
-        let md = archive_dir.join("conversation.md");
-        if !md.exists() {
-            anyhow::bail!(
-                "No transcript found in archive (expected session.jsonl or conversation.md)"
-            );
-        }
-        (fs::read_to_string(&md)?, true)
-    };
+    let source_file = archive_source(&archive_dir).ok_or_else(|| {
+        anyhow::anyhow!(
+            "No transcript found in archive (expected conversation.md or session.jsonl)"
+        )
+    })?;
+    let is_clean_md = source_file.extension().and_then(|e| e.to_str()) == Some("md");
+    let content = fs::read_to_string(&source_file)?;
 
     if let Some(pattern) = grep_pattern {
         // Filter lines matching pattern — works identically on both formats
@@ -372,7 +367,10 @@ pub fn read_session(
         print!("{}", content);
     }
 
-    // Include agent transcripts if requested
+    // Include agent transcripts if requested.
+    // Agent transcripts are always stored as JSONL (even in clean-mode archives
+    // that use conversation.md for the main session), so we only look for .jsonl
+    // files in the agents/ directory.
     if include_agents {
         let agents_dir = archive_dir.join("agents");
         if agents_dir.exists() {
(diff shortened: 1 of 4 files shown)


coryzibell (changes requested):
## Review — PR #224: Codex archive format detection + graph observability endpoints

Acknowledgments first. The `archive_source()` helper and its tests are clean — the preference-order test covering all four states (md-only, both, jsonl-only, neither) is exactly right. The `search_archives` refactor with proper error handling and the `skipped` counter is a real improvement over the silent `if let Ok(...)` swallowing. The observability endpoints are useful features and the overall structure is sound.

That said, the squirrels found issues. Organized by severity.

---

## Must-Fix

**1. Format preference inconsistency between `read_session` and `archive_source`.**

`src/codex.rs` — `read_session` checks `session.jsonl` first, then falls back to `conversation.md`. `archive_source` does the opposite: `conversation.md` first, then `session.jsonl`. When both files exist in an archive directory, `read_session` reads the JSONL while `search_archives` (via `archive_source`) reads the markdown. Two commands hitting the same archive return results from different files.

`read_session` should call `archive_source()` instead of reimplementing its own detection with inverted priority. Since clean-mode is the post-March-31 default, md-first is the correct order, and `archive_source` already has it right.

**2. Conditional `count()` in SurrealQL — verify against actual DB version.**

`src/surreal_db.rs` — The `graph_health_async` query uses `count(embedding IS NOT NONE)` and similar conditional count expressions inside `GROUP ALL`. This syntax is version-dependent in SurrealDB. Some versions require `math::sum(IF condition THEN 1 ELSE 0 END)` instead. This needs verification against v2.6.1 specifically, or a fallback approach that works across versions. A query that silently returns 0 for all conditional counts would produce plausible-looking but completely wrong health data.

**3. Null `last_activated` in stale query.**

`src/surreal_db.rs` — The stale-high-res condition `last_activated < time::now() - duration::from::days(30) AND resonance >= 5` does not handle null. Entries that have never been activated will have `last_activated` as NONE. Whether `NONE < some_datetime` evaluates to true or false depends on SurrealDB's null comparison semantics, and either way the result is wrong: these entries are either silently excluded (under-counting stale entries) or silently included (correct by accident, but fragile).

Should be: `(last_activated IS NONE OR last_activated < time::now() - duration::from::days(30)) AND resonance >= 5`

---

## Should-Fix

**4. `include_agents` / search path only looks for `.jsonl` — misses clean-mode agent transcripts.**

If there are other callers or downstream consumers that enumerate archive contents by extension, they will have the same blind spot that `search_archives` had before this PR. At minimum, add a comment in or near `archive_source` documenting the agent transcript format assumptions so the next person knows where the boundary is.

**5. Duplicated `surreal_path` / DB-open boilerplate in `src/main.rs`.**

All three new handler arms (`Health`, `Growth`, `OpenThreads`) repeat the same two lines:
```rust
let surreal_path = config.db_path.with_extension("surreal");
let db = crate::surreal_db::SurrealDatabase::open_with_verbose(surreal_path, verbose)?;
```
Extract a helper or hoist the DB open above the match arm. Three copies is a maintenance hazard — the fourth copy will get it wrong.

**6. `decay_score` gives maximum score to unparseable `created_at`.**

`src/surreal_db.rs` — When `DateTime::parse_from_rfc3339` fails, `created_secs` is set to `0.0`. Then the `weeks` calculation hits the `else` branch where `created_secs` is not `> 0.0`, so `weeks` is set to `0.0`. That means no decay is applied — `score = resonance * 0.95^0 = resonance`. An entry with a garbage date floats to the top of the open-threads list. Should treat unparseable dates as maximally old (e.g., set `weeks` to some large value like `520.0` — ten years), not maximally fresh.

**7. `open_threads_async` double-filters on state.**

`src/surreal_db.rs` — The SurrealQL WHERE clause filters on `summary IS NONE OR summary.state IS NONE OR summary.state = 'open'`. Then the Rust `filter_map` parses `summary` again and filters on `state != "open"`. The concern is that `summary` is parsed differently in each layer — SurrealDB treats it as a DB field, Rust re-parses it as JSON string or object. If SurrealDB's object access and Rust's `serde_json::from_str` disagree on the shape, the two filters will disagree on which entries are open. Pick one source of truth. The DB-side filter is the right one; the Rust side should trust it and only do reshaping.

**8. Week-bucket math is epoch-relative.**

`src/surreal_db.rs` — `unix_seconds / 604800` produces week numbers whose boundaries fall on Thursday 00:00:00 UTC (since the Unix epoch started on a Thursday). The 56-day cutoff in the SurrealQL WHERE clause won't align precisely with the 8-bucket boundary computed in Rust. Minor data discrepancy at the edges. Worth a one-line comment documenting the Thursday-boundary behavior so no one wastes time debugging off-by-one sparkline values later.

---

Fix items 1-3 before merge. Items 4-8 are strong recommendations.

PR #259, opened 2026-04-27
refactor: align all filesystem paths to canonical $MX_HOME via src/paths.rs

coryzibell:
Closes #255

## Summary

Pulls every filesystem path in `mx` through the canonical `src/paths.rs`
module. Implements all nine decisions from #255, plus the Acceptance
criteria.

- **Zero `dirs::home_dir()` calls outside `src/paths.rs`** (verified)
- **Zero hardcoded `~/`, `.mx`, `.crewu`, `/etc/mx` strings outside
  `src/paths.rs`** (verified)
- New `paths.rs` helpers for kv, state, memory, fastembed, and
  external Claude data
- `paths::agents_dir()` and `paths::schemas_dir()` removed
- `MX_MEMORY_PATH` removed (renamed to `MX_SURREAL_ROOT`); old name
  triggers a one-line stderr note
- `MX_STATE_SCHEMA` removed (replaced by `--schema {id|path}` CLI flag)
- New `MX_ISOLATE_FASTEMBED` flag for mx-owned model cache
- Legacy `StateSchema` (and `src/state/schema.rs`, `src/state/loader.rs`)
  deleted
- `mx memory rebuild` removed; `KnowledgeEntry::from_markdown` and
  `rebuild_index` deleted as dead code
- `mx agents seed` -> `mx memory seed agents` (dir `$MX_HOME/agents/`
  -> `$MX_HOME/memory/seed/agents/`)
- `mx memory import` -> `mx memory seed knowledge` (multi-file scan
  at `$MX_HOME/memory/seed/knowledge/*.jsonl`)
- All renamed/removed commands have clap-hidden bail stubs that point
  to the new spelling -- friendlier than "unknown subcommand"
- Soft fallback + stderr warning + `TODO(<name>-migration)` comment on
  every path migration (kv, agents seed, knowledge seed)
- README env-var table + layout diagram updated

## Final layout

```
~/.mx/
├── kv/{schema,data}/{agent}.{toml,json}
├── state/schemas/{id}.yaml          # default ID: "tensor"
├── memory/
│   ├── surreal/                     # MX_SURREAL_ROOT
│   ├── embed/                       # MX_ISOLATE_FASTEMBED only
│   └── seed/{agents,knowledge}/
├── codex/                           # MX_CODEX_PATH
├── cache/sync/{owner-repo}/
├── artifacts/
└── swap/
```

## Test results

- `cargo build --release` -- clean
- `cargo test --all-features` -- **477 passed, 0 failed, 3 ignored**
[...]

PR #261, opened 2026-04-27
docs: add comprehensive filesystem layout & env var reference

coryzibell:
## Summary

Adds a comprehensive user-facing reference for `mx`'s filesystem layout,
environment variables, and the conventions that govern them. Slims the
README's Configuration section to a teaser pointing at the new doc.

Follow-up to #259 (the path-alignment refactor) -- everything that
landed there is now documented in one canonical place.

## What's new

`docs/paths.md` (288 lines) covering:

1. **The principle** -- single base directory, per-subsystem subdirs,
   override precedence (per-file > per-subsystem > `MX_HOME` > default)
2. **Layout** -- the full tree with a paragraph per subdirectory
   (kv/, state/, memory/, codex/, cache/, artifacts/, swap/) explaining
   what command(s) use it and what file shapes live there
3. **Environment variables** -- five themed tables with anchor links
   (path overrides, SurrealDB connection, GitHub App auth, identity &
   display, tuning) plus an explicit "Removed" section for
   `MX_MEMORY_PATH` and `MX_STATE_SCHEMA`
4. **Migration & legacy fallbacks** -- table of legacy locations, new
   locations, and behavior; explicit "no data is lost" reassurance
5. **Renamed and removed CLI commands** -- old/new mapping table for
   `mx agents seed` -> `mx memory seed agents`, `mx memory import` ->
   `mx memory seed knowledge`, `mx memory rebuild` -> removed,
   `MX_STATE_SCHEMA` -> `--schema` flag
6. **Examples** -- 7 worked scenarios (move the tree, SSD split, KV
   templating, FastEmbed isolation, schema flag, network SurrealDB)
7. **Notes for contributors** -- the `paths.rs` rule, in two
   paragraphs, with the `_with` test-seam pattern documented

## What changed in README

Configuration section trimmed from ~50 lines to ~18: a 5-paragraph
orientation, a 2-bullet "removed" callout for the most disruptive
breaking changes, and a bold link to `docs/paths.md`. README total
dropped from 197 to 172 lines. No duplicated env-var info that could
drift.

## Test plan

- [x] Markdown renders correctly (eyeballed for fenced-code-block balance)
- [x] All file paths in the doc match the actual code (`src/paths.rs` exists, helpers like `kv_schema_path` are real)
- [x] Anchor links resolve within the doc
- [x] No duplicated env-var documentation between README and docs/paths.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,46 +144,22 @@ Use `mx log` to read commits. Use `git log` if you enjoy puzzles.
 
 ## Configuration
 
-Filesystem layout (everything lives under `$MX_HOME`, default `~/.mx/`):
+Everything mx writes lives under a single base directory: `$MX_HOME`, default
+`~/.mx/`. Each subsystem owns a subdirectory (`kv/`, `state/`, `memory/`,
+`codex/`, ...). Move the whole tree by setting `MX_HOME`, or override one
+subsystem at a time with vars like `MX_SURREAL_ROOT`, `MX_CODEX_PATH`,
+`MX_KV_SCHEMA`, `MX_KV_DATA`, `MX_ISOLATE_FASTEMBED`.
 
-```
-~/.mx/
-├── kv/
-│   ├── schema/{agent}.toml     # override MX_KV_SCHEMA
-│   └── data/{agent}.json       # override MX_KV_DATA
-├── state/
-│   └── schemas/{id}.yaml       # default ID: "tensor"; CLI --schema flag
-├── memory/
-│   ├── surreal/                # override MX_SURREAL_ROOT
-│   ├── embed/                  # only when MX_ISOLATE_FASTEMBED is set
-│   └── seed/
-│       ├── agents/             # for `mx memory seed agents`
-│       └── knowledge/          # for `mx memory seed knowledge`
-├── codex/                      # override MX_CODEX_PATH
-├── cache/sync/{owner-repo}/
-├── artifacts/
-└── swap/
-```
+Two env vars were removed in the path-alignment refactor (#259):
 
-Key environment variables:
+- `MX_MEMORY_PATH` -- use `MX_SURREAL_ROOT` instead. Setting the old name now
+  emits a one-line stderr note and is otherwise ignored.
+- `MX_STATE_SCHEMA` -- replaced by the `mx state ... --schema {id|path}` CLI
+  flag. The default schema ID is now `tensor` (was `crewu`).
 
-| Variable | Purpose |
-|----------|---------|
-| `MX_HOME` | Base directory (default `~/.mx/`) -- the only filesystem root |
-| `MX_CURRENT_AGENT` | Active agent identity (required for `memory wake`, used as default for `--source-agent`) |
-| `MX_SURREAL_ROOT` | Override SurrealDB storage root (default `$MX_HOME/memory/surreal/`) |
-| `MX_SURREAL_MODE` | SurrealDB connection mode (`embedded` or `network`) |
-| `MX_CODEX_PATH` | Override codex archive storage path (default `$MX_HOME/codex/`) |
-| `MX_KV_SCHEMA` | Override kv schema path (supports `{agent}` placeholder) |
-| `MX_KV_DATA` | Override kv data path (supports `{agent}` placeholder) |
-| `MX_ISOLATE_FASTEMBED` | When set, store fastembed model cache under `$MX_HOME/memory/embed/` instead of the shared XDG cache |
-| `MX_USER_NAME` | Display name for user in codex transcripts |
-| `MX_ASSISTANT_NAME` | Display name for assistant in codex transcripts |
-
-Removed in favor of CLI flags / renames:
-
-- `MX_MEMORY_PATH` -- renamed to `MX_SURREAL_ROOT`. Setting the old variable now emits a one-line stderr note (will be dropped after one release cycle).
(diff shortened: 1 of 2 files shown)


coryzibell:
## Code Review (docs)

**REQUEST CHANGES** -- one factual error in the identity table; otherwise accurate but with a few warnings worth fixing before this becomes the go-to reference at 2am.

### C - Factual errors

- **`docs/paths.md` -> Identity & display table, `MX_ASSISTANT_NAME` row.** Default is listed as "derived." Per `src/codex/transcript.rs:71`, the fallback is the literal string `"Orchestrator"` -- nothing is derived. `MX_USER_NAME` *is* derived (env > git `user.name` > `"User"`). Split the row: `MX_USER_NAME` = "derived (git `user.name`, else `User`)"; `MX_ASSISTANT_NAME` = `Orchestrator`.

- **`docs/paths.md` -> Layout tree, `state/schemas/{id}.yaml` line.** Tree comment says "default ID: tensor; CLI --schema flag" but the doc never tells the reader the file extension constraint encoded in `tensor_schema_path` (`paths.rs:149`): the helper *only* produces `.yaml` paths. JSON schemas work only via the explicit `--schema /path/to/x.json` form, never via bare ID. Worth one sentence under `state/schemas/`, since the env-paths section claims `--schema` accepts `.json`.

### W - Significant gaps / claims that could mislead

- **`docs/paths.md` -> Examples block, `mx state encode --schema crewu ...`.** This example will fail on a fresh install -- the default schema ID was renamed to `tensor` and there is no `crewu.yaml` shipped under `state/schemas/`. Either change the example to `--schema tensor`, drop the `--schema` flag entirely (defaults already resolve), or note that `crewu` here assumes the user authored their own schema file.

- **`docs/paths.md` -> Migration & legacy fallbacks, knowledge-seed row.** The legacy path is given as `$MX_HOME/memory/index.jsonl` -- correct -- but `legacy_memory_index_jsonl()` in `paths.rs:201` returns a single file, while the new location is a *directory* of `*.jsonl` files. Worth one clause: "the old location was a single hardcoded file; the new location is a directory scanned for every `*.jsonl`." A reader migrating mid-shift will assume rename, not shape change.

- **`docs/paths.md` -> "Empty-string env values are treated as unset."** True for `MX_HOME`, `MX_SURREAL_ROOT`, `MX_CODEX_PATH`, `MX_KV_*`, `MX_ISOLATE_FASTEMBED`. **Not true** for the SurrealDB connection vars in `surreal_db/connection.rs::from_env` -- e.g. `MX_SURREAL_USER=""` produces user `""`, not `root`. (Password is filtered by `.filter(|p| !p.is_empty())`; the rest are not.) Either narrow the claim ("Empty-string values for path overrides are treated as unset") or fix the SurrealDB resolvers.

### S - Improvements

- **`docs/paths.md` -> `state/schemas/` section.** Add the path-vs-ID heuristic from `handlers/state.rs:14-21` verbatim or near-verbatim. The current doc says "contains a slash or ends in `.yaml`/`.yml`/`.json`" but omits the gotcha that `--schema my.schema` is treated as an ID, not a relative path. That edge case is exactly the kind of thing the 2am reader hits.

- **`docs/paths.md` -> Notes for contributors.** "Do not call `dirs::home_dir()` outside `src/paths.rs`" is stated as a rule, but `paths.rs` itself calls it in `claude_projects_dir`/`claude_config_path` (i.e. external read-only sources). Add the carve-out so a future contributor doesn't move those out "for consistency."

### N - Polish

- "Subsystem" / "module" / "subdirectory" used interchangeably. Pick one. The principle section says "subsystem"; the layout paragraphs slip into "directory."
- Removed table -> `MX_STATE_SCHEMA` row has unescaped `|` inside `{id\|path}`. Renders fine in GFM but inconsistent with the same token elsewhere -- standardize.

---

Fix the C items, address the empty-env claim and the `crewu` example, and the rest is acceptable.

coryzibell:
## Round 2 fixes pushed

Three commits since the last review:

- `abe3853` -- addresses Verdictia's C1, C2, W1, W2, W3 (the original factual errors and clarifications)
- `0d290e7` -- addresses S2 (contributor notes carve-out)
- `f85c315` -- **corrects two factual errors introduced by `abe3853`**:
  - `state/schemas/` bare-ID lookup: previously claimed `.yml` was not tried for bare-ID lookup. Code at `tensor.rs::load_by_id` (lines 113-130) tries `.yaml`, then `.yml`, then `.json`. Doc now lists all three in order.
  - Path-vs-ID heuristic: previously claimed any dot flips classification to path. Actual code at `handlers/state.rs:24-32` checks for slash OR recognized extension only. Doc now states the rule accurately. Example fixed: `--schema my.schema` is an ID lookup, not a path.

Spot-checked against the actual code lines. Ready for focused re-review on the `state/schemas/` section and any compounding effects.

coryzibell:
## Final fix pushed

`b63e4fd` -- corrects the contributor-notes carve-out flagged in the previous re-review.

**Before:** "today, only to compute the `~/.mx/` default for `mx_home()`. If a future helper needs to point at a path owned by another tool..."

**After:** lists all five actual callers of `dirs::home_dir()` in `paths.rs`:
- `mx_home()` (~/.mx/ default)
- `legacy_crewu_kv_schema_path`, `legacy_crewu_kv_data_path` (kv migration legacy fallbacks)
- `claude_projects_dir`, `claude_config_path` (read-only Claude data)

Verified against `paths.rs:29, 121, 132, 264, 272`.

PR #265, opened 2026-04-28
feat(state): ship default tensor.yaml schema, self-seed on first run

### LINKED ISSUE #256: ship a generic tensor.yaml default schema (derive from crewu, strip harness-specific moods)
## Context

Follow-up to #255 (path alignment work). That issue moves TensorSchema
files to `$MX_HOME/state/schemas/{id}.yaml` and changes the default
schema ID from `crewu` to `tensor`. This issue ships the actual
`tensor.yaml` default schema file.

## What

Create a generic `tensor.yaml` schema and ship it as the default
public-facing tensor state schema for `mx state`.

## Why

`crewu.yaml` -- the current default -- is the schema for a specific
agent harness. Most of its dimensions (temperature, entropy, agency,
connection, weight) are generally useful, but the named *moods* are
harness-specific landmarks that won't mean anything to a public mx user.

Default schemas should work out of the box for anyone running `mx state`
without prior context.

## Approach

1. Start from `crewu.yaml` (the proven base)
2. Keep the five dimensions and their anchors (these are general-purpose)
3. Strip or generalize the named moods -- replace harness-specific
   landmarks with a small set of universally-recognizable mood names
   (or ship with no moods and let users add their own)
4. Ship as `artifacts/etc/state/schemas/tensor.yaml` (or similar staging
   location), copied to `$MX_HOME/state/schemas/tensor.yaml` on first
   `mx state` invocation if missing

## Open questions

- Where does the seed file live in the repo before install/copy?
- Should `mx state` self-seed on first run, or should there be a
  `mx state schemas init` command that does it explicitly?
- Should `crewu.yaml` continue to ship alongside as an example, or
  is it removed from the public artifacts entirely?

## Depends on

- #255 -- path alignment and TensorSchema relocation must land first

coryzibell:
Closes #256

## Summary

Ships a public-friendly default tensor schema (`tensor.yaml`) and
self-seeds it on first invocation of `mx state`. New installs work
without any manual setup; existing installs with a user-authored
`tensor.yaml` are left untouched.

The schema has six dimensions tuned for general-purpose Claude usage:
`entropy`, `agency`, `temperature`, `verbosity`, `skepticism`, `humor`.
No `moods` block -- those are user-authored landmarks.

## Implementation

- New file `schema/state/schemas/tensor.yaml` -- the canonical default
  (six dimensions; defaults nudged off-center for `verbosity` and
  `skepticism` to compensate for known LLM biases)
- Embedded into the binary via `include_str!`
- New `ensure_default_schema_seeded()` in `src/tensor.rs` writes the
  embedded content to `$MX_HOME/state/schemas/tensor.yaml` if-and-only-if
  no file exists at that path
- Wired into `TensorSchema::load_by_id` for `schema_id == "tensor"` only;
  `load_default()` flows through the same path
- Test seam: `ensure_schema_seeded_at(path, content) -> Result<bool>`
  follows the existing `_with(env_val, home)` pattern used elsewhere in
  the codebase (no `MX_HOME` mutation in tests)

## Tests

5 new tests:
- Fresh install seeds the file and `load_default` succeeds end-to-end
- Second invocation is idempotent (file mtime unchanged)
- User-authored file at the path is preserved (embedded version not
  written)
- Embedded YAML parses cleanly via the existing `TensorSchema` loader
- Asset location parity (`schema/state/schemas/tensor.yaml` matches
  what gets seeded at runtime)

## Test results

- `cargo build --release` -- clean
- `cargo test --all-features` -- **519 passed, 0 failed, 3 ignored** (was 514)
- `cargo clippy --all-features -- -D warnings` -- clean
- `cargo fmt -- --check` -- clean

## Runtime verification

`MX_HOME=$(mktemp -d) cargo run --release -- state encode --dimensions "..."`
on a fresh tempdir:
- Command succeeds
- `$MX_HOME/state/schemas/tensor.yaml` is written
- `diff` against the embedded source: clean
- Second invocation: file mtime unchanged (idempotent in production)

## Asset location decision

Placed at `schema/state/schemas/tensor.yaml` rather than under a new
top-level `assets/` directory. Matches the existing `schema/` repo
convention (neighbor: `schema/surrealdb-schema.surql`) and mirrors the
runtime layout under `$MX_HOME/state/schemas/`.

## Test plan

- [x] All cargo gates green
- [x] Fresh install seeds and works (runtime-verified)
- [x] Idempotent (test + runtime check)
- [x] User-authored file preserved (test)
- [ ] Manual: confirm visual `mx state show` output looks reasonable
      with the new defaults

### DIFF
--- a/schema/state/schemas/tensor.yaml
+++ b/schema/state/schemas/tensor.yaml
@@ -1,0 +1,63 @@
+# Default Tensor Schema
+#
+# Six-dimension default schema for `mx state` encoding. This file is
+# embedded into the binary via `include_str!` and self-seeded into
+# `$MX_HOME/state/schemas/tensor.yaml` on first run when no user-authored
+# version is present.
+#
+# Defaults are nudged off-center where the underlying model's natural bias
+# warrants compensation. Edit your local copy under `$MX_HOME` to override;
+# the seeder will not overwrite an existing file.
+
+id: tensor
+name: Default Tensor
+version: 1
+
+dimensions:
+  - id: entropy
+    name: Entropy
+    anchors:
+      low:  ordered / focused / coherent
+      mid:  structured but breathing
+      high: chaotic / associative / wild
+    default: 0.4
+
+  - id: agency
+    name: Agency
+    anchors:
+      low:  receptive / yielding / responsive
+      mid:  collaborative
+      high: active / driving / proactive
+    default: 0.5
+
+  - id: temperature
+    name: Temperature
+    anchors:
+      low:  cold / precise / formal
+      mid:  measured warmth
+      high: warm / casual / unguarded
+    default: 0.5
+
+  - id: verbosity
+    name: Verbosity
+    anchors:
+      low:  terse / minimal / no waste
+      mid:  balanced
+      high: expansive / elaborate / generous
+    default: 0.4
+
+  - id: skepticism
+    name: Skepticism
+    anchors:
+      low:  agreeable / affirming / supportive
+      mid:  helpful but honest
+      high: challenging / pushback / red-team
+    default: 0.55
+
+  - id: humor
+    name: Humor
+    anchors:
+      low:  serious / matter-of-fact
+      mid:  occasional dry wit
+      high: playful / quippy / TARS at 90%
+    default: 0.3
--- a/src/tensor.rs
+++ b/src/tensor.rs
@@ -3,12 +3,11 @@
 //! Implements the tensor encoding system designed by Schemnya.
 //! Values-first: encode dimensional values -> derive nearest mood label.
 //!
-//! The crewu schema defines 5 dimensions:
-//! - temperature (ᚣ): cold/precise <-> warm/playful
-//! - entropy (ᚤ): ordered/focused <-> chaotic/wild
-//! - agency (ᚡ): receptive/yielding <-> active/driving
-//! - connection (ᚢ): distant/separate <-> close/merged
-//! - weight (ᚠ): light/floating <-> heavy/grounded
+//! Schemas are user-authored YAML files; the default `tensor` schema ships
+//! with six dimensions (entropy, agency, temperature, verbosity, skepticism,
+//! humor) and self-seeds at `$MX_HOME/state/schemas/tensor.yaml` on first
+//! `mx state` invocation. See `schema/state/schemas/tensor.yaml` for the
+//! shipped content.
 
 use anyhow::{Context, Result, bail};
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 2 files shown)


coryzibell:
## Code Review

**APPROVE WITH FIXES**

PR #265 cleared all three reads. Schema content matches the architecture comment byte-for-byte. Self-seed logic is clean: pure test seam, production wrapper routes through `paths::tensor_schema_path`, parent dirs created via `create_dir_all`, gated on `schema_id == "tensor"` only, idempotent on second run. All five new tests do what they claim.

### Warnings

**W1 - Stale module doc comment.** `src/tensor.rs:6-11` still reads:

> The crewu schema defines 5 dimensions: temperature, entropy, agency, connection, weight

This is no longer the default. The shipped default is the 6-dim `tensor` schema (entropy, agency, temperature, verbosity, skepticism, humor). The very first thing a reader sees when opening this file now lies to them. Replace with a description of the 6-dim default, or rewrite as "Schemas are user-authored YAML; the default `tensor` schema ships with six dimensions (see `schema/state/schemas/tensor.yaml`)."

### Suggestions

**S1 - Embed/disk parity is implicit, not explicit.** PR description claims a parity test exists. I count five new tests; none does an explicit "read disk file, compare to `DEFAULT_TENSOR_SCHEMA_YAML`" assertion. The `fresh_install_seed_writes_embedded_content` test compares on-disk-after-seed to the embed -- but since the embed is `include_str!` of that exact file, this is tautological at the build level. Acceptable in practice (cargo build would fail if the file moved), but the PR description overclaims. Either add an explicit `assert_eq!(fs::read_to_string("schema/state/schemas/tensor.yaml")?, DEFAULT_TENSOR_SCHEMA_YAML)` or update the PR description.

**S2 - `ensure_default_schema_seeded` discards the `bool`.** The wrapper throws away the `Result<bool>` from the seam. Fine for current callers, but if a future caller wants to log "first-run seed performed" the production wrapper would need re-plumbing. Consider returning `Result<bool>` here too -- zero cost, future-proof.

### Notes

**N1 - TOCTOU race on simultaneous first-run `mx state` invocations.** `path.exists()` -> `fs::write` is non-atomic. Two concurrent first-runs both see `false`, both write. Content is identical so the result is correct; not worth fixing.

**N2 - No `crewu.yaml` shipped.** Verified: `schema/state/schemas/` contains only `tensor.yaml`. Matches architecture decision.

Fix W1. S1 and S2 at implementer's discretion.

coryzibell:
## Code Review (re-review)

**APPROVE** -- ship it

All three prior findings resolved cleanly. New test fires. No new behavior introduced beyond the three fixes. No regressions.

### Per-finding verdict

- **W1 module doc** -- resolved. `src/tensor.rs:6-11` now describes the 6-dim default with the correct dimension names: entropy, agency, temperature, verbosity, skepticism, humor. No "crewu" / "5 dimensions" residue.
- **S1 embed/disk parity test** -- resolved. `embedded_yaml_matches_disk_source` at `src/tensor.rs:638-654` reads `schema/state/schemas/tensor.yaml` via `std::fs::read_to_string` and asserts equality with `DEFAULT_TENSOR_SCHEMA_YAML`. Path resolves relative to crate root; test executed and passed.
- **S2 plumbed bool** -- resolved. `ensure_default_schema_seeded` at `src/tensor.rs:59-62` returns `Result<bool>`. Call site at `src/tensor.rs:162` consumes via `let _seeded = ensure_default_schema_seeded()?;` -- explicit, named, propagates error, ignores bool deliberately. Clean.

### Scan for new breakage

- Diff scope matches the three fixes. No extraneous changes.
- Parity test runs in 0.00s and passes. Total test count consistent with +1.
- Type signature change on `ensure_default_schema_seeded` is internal (private fn); single call site updated.
- Pre-existing `surreal_db::tests::test_open_with_path` flake noted -- confirmed not in scope.

The verdict is final. Ship.

PR #267, opened 2026-04-30
refactor(codex): paths.rs helpers + index/ skeleton + manifest v2→v5 (PR 1/5 of #254)

coryzibell:
**Part of:** #254

**Architecture context:** https://github.com/coryzibell/mx/issues/254#issuecomment-4349828527

## Summary

Foundation PR for the codex unification series (PR 1 of 5). Adds new path helpers in `src/paths.rs` (no callers yet — PR 2 wires them), the `src/codex/index/` module skeleton with `ProjectIndex` / `ProjectEntry` / `IndexError` types behind `unimplemented!()` stubs (PR 2 implements writes; PR 3 implements reads), and bumps the manifest write version from 2 to 5 with optional new fields (`tool_output_count`, `mcp_log_count`, `history_lines`, `source_breakdown`).

The version bump is intentional — v5 manifests from this PR contain the same bytes as v2 plus a higher version number, since archive does not yet write the new sidecars.

## Files affected

- `src/paths.rs` — +7 helpers, +7 unit tests
- `src/codex/index/mod.rs` — new file: `ProjectIndex`, `ProjectEntry`, `IndexError`, 3 smoke tests
- `src/codex/mod.rs` — new optional fields + `SourceBreakdown` struct + `pub mod index;`
- `src/codex/archive.rs` — write version 2 → 5 at two sites; uses new `MANIFEST_WRITE_VERSION` constant
- `src/codex/migrate.rs` — v2 → v5 metadata-only migration path; pairwise to match existing v1 → v2 image-extraction migration shape

## Behavior

**No behavior change.** No new CLI surface. No new flags. Existing `mx codex archive` and `mx codex migrate` invocations produce identical outputs (modulo the version field being `5` instead of `2`).

## Test plan

- [x] `cargo build` clean
- [x] `cargo test` — 533 passed / 0 failed / 3 ignored (+13 from the 520 baseline on main)
- [x] `cargo clippy --all-targets` clean
- [x] v2 → v5 round-trip test confirms existing v2 manifests still deserialize and re-serialize as v5 with new fields elided

## Notes

- The known `surreal_db::tests::test_open_with_path` flake did not fire this run (now broken twice in three loopas — possibly luck, possibly latent fix; not in scope here).
- Two minor parameterization decisions documented in the report: `claude_subagents_dir(project_slug, session_id)` and `tmp_claude_tasks_dir(uid, user_slug, session_uuid)` — both took the multi-arg form because path components beyond the entry point are required to build the path.

Issue #254, opened 2026-04-27
combine mx codex archive and mx session export -- overlapping ingestion of Claude sessions


coryzibell:
## What's overlapping

Two commands ingest the same Claude session source and produce
overlapping outputs:

- `mx codex archive [--all]` (`src/codex/archive.rs`) -- scans Claude
  sessions and archives them into the codex
- `mx session export` (`src/session.rs`) -- reads a Claude session and
  exports it as a transcript

Both walk the same source tree. Both transform JSONL conversations into
a persisted, queryable form. The split is more accident-of-history than
deliberate design.

## What "combined" might look like

Open question for the implementer -- some shapes worth considering:

- One command, two output modes (e.g. `mx archive [--format codex|transcript]`)
- One command writes both forms in one pass (codex is the structured store,
  transcript is a derived view)
- Keep both as user-facing names but route through a single shared
  ingestion module

## Why now

Today, fixing a bug in session ingestion (filtering, parsing, source
discovery) requires touching both modules. Filing this so the unification
gets considered before either grows further.

## Out of scope

Just observation + design discussion. No implementation proposal yet.

coryzibell:
## Proposed Architecture

*Synthesized from a sprawl audit — full audit available in shift notes; key sources mapped: `~/.claude/projects/`, `~/.claude/projects/.../subagents/`, `~/.claude/sessions/`, `~/.claude/history.jsonl`, `~/.cache/claude-cli-nodejs/.../mcp-logs-*/`, `/tmp/claude-<uid>/.../tasks/*.output`, `~/.wonka/vault/archives/`.*

### 1. The vision (one line)

**Issue's option 3, sharpened.** Keep `mx codex` and `mx session` as user-facing names, but route both through a new `src/ingest/` module that owns source enumeration, parsing, and emission — with `codex/archive.rs` and `session.rs` becoming ~30-line CLI handlers that pick an emitter. Rejected option 1 (`mx archive --format`) because it forces a breaking rename to win a single new flag, and rejected option 2 (one pass writes both forms) because most invocations want exactly one output and the dual-write tax is paid every time.

### 2. The source layer (tiered)

**Tier 1 — must include in v1:**
- `~/.claude/projects/<project>/<uuid>.jsonl` — both commands already read this.
- `~/.claude/projects/<project>/<uuid>/subagents/agent-*.jsonl` — `archive.rs:261-309` already walks these; `session.rs:169-173` silently drops them. Closing this gap is the headline win of unification.
[...]

PR #268, opened 2026-04-30
refactor(codex): archive skeleton + new source walkers + index writes (PR 2/5 of #254)

coryzibell:
**Part of:** #254 (PR 2 of 5 — structural; only PR 5 closes the umbrella)

**Architecture context:** https://github.com/coryzibell/mx/issues/254#issuecomment-4349828527

**Prior PR:** #267 (foundation — `paths.rs` helpers, index skeleton, manifest v5 bump)

## Summary

Restructures `src/codex/archive.rs` into a directory module (`archive/{mod,write,sources,paths,include}.rs`) behind a new `archive::run(ArchiveRequest, ArchiveOptions) -> ArchiveResult` entry point. Adds three new optional source walkers (MCP logs, /tmp tool outputs, history-jsonl slice) gated behind a new `--include` flag with status-quo default (`subagents` only). Implements `ProjectIndex::open` and `rebuild_from_manifests` with symlink-format index storage, atomic-ish staging-then-rename swap; wired into `archive::run` as best-effort post-write step. Status-quo invocations produce **byte-identical manifests** to today (asserted by a new test).

## Files affected

- **Restructure:** `src/codex/archive.rs` deleted; new `src/codex/archive/{mod,write,sources,paths,include}.rs`
- **CLI wiring:** `src/cli.rs` (lines 1190-1206), `src/handlers/mod.rs` (lines 88-99)
- **Schema population:** `src/codex/mod.rs` (no struct changes; only writer-side population)
- **Index implementation:** `src/codex/index/mod.rs` (`open`/`open_under`/`rebuild_from_manifests` implemented; `lookup`/`is_stale` still stubbed for PR 3)

## Behavior

**No CLI behavior change for status-quo invocations.** Existing `mx codex archive` and `mx codex archive --all` produce byte-identical manifest output. New capability ships as opt-in `--include` values: `mcp`, `tool-output`, `history`. The `subagents` token is on by default, matching today's behavior.

**One judgment call worth noting:** the existing `--include-agents` flag was NOT aliased to `--include subagents`. They control different axes — `--include-agents` is `requires = "clean"` and controls transcript inclusion in `conversation.md`, while `--include subagents` controls source-file capture. Aliasing would conflate two orthogonal axes. Help text on `--include` documents the distinction.

## Test plan

- `cargo build` clean
- `cargo test` — 565 passed / 0 failed / 3 ignored (+32 from the 533 baseline on main post-PR-#267)
- New test `archive/mod.rs::status_quo_manifest_omits_new_v5_fields` runs the writer end-to-end against a tempdir codex and asserts none of `tool_output_count`, `mcp_log_count`, `history_lines`, `source_breakdown` appear in the serialized manifest under status-quo `IncludeSet`
- +32 new tests across paths (4), `IncludeSet` (12), source walkers (5), status-quo manifest (1), index (10)

## Notes

- Index format: relative symlinks (`by-project/<slug>/<archive-dir-name> -> ../../<archive-dir-name>`). Codex tree is movable as a unit. Atomic-ish swap via `by-project.staging/` → rename to bound partial-state risk.
- Index honors `MX_CODEX_PATH` (lives at `<codex_dir>/by-project/`).
- One transient test failure observed during development that did not reproduce — consistent with the known `surreal_db::tests::test_open_with_path` flake (now in its 5th confirmed shift; ticket overdue, not in scope here).

## What this PR does NOT do

- Does NOT add `mx codex export` (PR 3)
- Does NOT change the default `IncludeSet` to capture all Tier 1 sources
- Does NOT implement `ProjectIndex::lookup` or `is_stale` (PR 3, when export reads the index)
- Does NOT add `--archive-first` (PR 3) or `--backfill` (PR 5)
- Does NOT touch `~/.wonka/bin/activate` (separate factory commit, post-series)

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1188,6 +1188,20 @@ pub enum CodexCommands {
         /// Include agent sub-session conversations in clean transcript
         #[arg(long, requires = "clean")]
         include_agents: bool,
+
+        /// Comma-separated list of optional source artifacts to capture.
+        ///
+        /// Recognized: `subagents`, `mcp`, `tool-output`, `history`, `all`,
+        /// `none`. Today's default behavior corresponds to `--include
+        /// subagents`. The other tokens enable forthcoming source walkers
+        /// (MCP server logs, /tmp tool outputs, history.jsonl slice).
+        ///
+        /// Note: this flag governs which source files are *captured* into
+        /// the archive sidecars. The separate `--include-agents` flag
+        /// controls whether subagent transcripts are folded into the
+        /// `conversation.md` rendering when `--clean` is set.
+        #[arg(long, default_value = "subagents")]
+        include: String,
     },
 
     /// List archived sessions
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1185,7 +1185,11 @@ pub enum CodexCommands {
         #[arg(long)]
         clean: bool,
 
-        /// Include agent sub-session conversations in clean transcript
+        /// Include agent sub-session conversations in clean transcript.
+        ///
+        /// Requires `subagents` in `--include` (the default; if you pass
+        /// `--include none` or any value without `subagents`, this flag
+        /// will error at parse time rather than silently no-op).
         #[arg(long, requires = "clean")]
         include_agents: bool,
 
--- a/src/codex/archive/include.rs
+++ b/src/codex/archive/include.rs
@@ -1,0 +1,188 @@
+//! Opt-in source selection for `mx codex archive`.
+//!
+//! `IncludeSet` controls which optional sidecars the writer captures.
+//! Today only `subagents` is on by default — the same artifact the
+//! pre-unification archive flow always copied. The other three flags
+//! (`mcp`, `tool_output`, `history`) opt in to the new walkers added
+//! in PR 2 and are off by default until export (PR 3) and the broader
+//! UX have stabilized.
+//!
+//! The set is parsed from a comma-separated CLI string (`--include
+//! subagents,mcp,history`). Two special tokens short-circuit:
+//! `all` enables every flag, `none` disables every flag.
+
+use anyhow::Result;
+
+/// Which optional source artifacts to capture during archive.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct IncludeSet {
+    /// Subagent JSONLs (`agents/`). Default: ON — matches the pre-PR-2
+    /// behavior, where the archive always copied them.
+    pub subagents: bool,
+    /// MCP server logs (`mcp/`). Default: OFF.
+    pub mcp: bool,
+    /// `/tmp/.../tasks/*.output` snapshots (`tool-output/`). Default: OFF.
+    pub tool_output: bool,
+    /// Sliced `~/.claude/history.jsonl` lines (`history/`). Default: OFF.
+    pub history: bool,
+}
+
+impl IncludeSet {
+    /// The set that reproduces today's `mx codex archive` defaults
+    /// byte-for-byte: subagents on, everything else off.
+    pub fn status_quo() -> Self {
+        Self {
+            subagents: true,
+            mcp: false,
+            tool_output: false,
+            history: false,
+        }
+    }
+
+    /// All four sources on.
+    pub fn all() -> Self {
+        Self {
+            subagents: true,
+            mcp: true,
+            tool_output: true,
+            history: true,
+        }
+    }
+
+    /// Nothing on. Useful for `--include none` and for tests that want a
+    /// minimum-noise archive (just session.jsonl + manifest).
+    pub fn none() -> Self {
+        Self::default()
+    }
+
+    /// Parse a comma-separated CLI value.
+    ///
+    /// Tokens are case-insensitive. Recognized: `subagents`, `mcp`,
+    /// `tool-output` (or `tool_output`), `history`, `all`, `none`.
+    /// Unknown tokens print a warning to stderr and are skipped — we
+    /// don't fail-hard so a future rename can land alongside an
+    /// older user script gracefully.
+    ///
+    /// `all` and `none` are exclusive overrides: if either appears
+    /// anywhere in the list, it wins for the whole field it touches
+    /// (`all` sets every flag on, `none` sets every flag off). When
+    /// they appear together, later tokens win — left-to-right reading
+    /// order, the same way a shell would interpret a chain of toggles.
+    pub fn parse(s: &str) -> Result<Self> {
+        let mut set = Self::default();
+        for raw in s.split(',') {
+            let token = raw.trim().to_ascii_lowercase();
+            if token.is_empty() {
+                continue;
+            }
+            match token.as_str() {
+                "all" => set = Self::all(),
+                "none" => set = Self::none(),
+                "subagents" => set.subagents = true,
+                "mcp" => set.mcp = true,
+                "tool-output" | "tool_output" => set.tool_output = true,
+                "history" => set.history = true,
+                other => {
+                    eprintln!(
+                        "warning: ignoring unknown --include token '{}' \
+                         (recognized: subagents, mcp, tool-output, history, all, none)",
+                        other
+                    );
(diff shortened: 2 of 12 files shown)


coryzibell:

### Critical

**C1. Status-quo byte-identity test does not actually prove byte-identity.** `archive/mod.rs::status_quo_manifest_omits_new_v5_fields` (lines 233-307) only asserts the *new field names* are absent from the serialized JSON. It does NOT compare against a golden pre-PR-2 manifest. The test would pass even if PR 2 silently reordered existing fields, renamed a field, or changed serialization (e.g. indentation). The "byte-identical" claim is not under test. Fix: golden-file fixture, modulo time-dependent fields, exercising the matrix (single+full, single+clean, all+full, all+clean). Today only single+full is exercised.

**C2. `session_start` derivation is wrong, and PR 2 now USES it as a real timestamp window.** `write.rs:290`: `let session_start: DateTime<Utc> = modified.into();` — that's the JSONL's `mtime`, not the session start. Before PR 2, this was cosmetic metadata. Now it's the lower bound of `TimestampWindow` (`write.rs:319`) feeding `find_mcp_logs` and `find_history_slice`. The window is therefore `[mtime, now]`. MCP logs and history lines from *before* the session ended will be excluded; lines written *after* mtime will be included. When `--include` flips on (PR 3 export), this will silently miss most of what users expect. Either (a) parse first/last event timestamps from the JSONL (the doc on line 317 already promises this), or (b) gate `mcp`/`history` behind a hidden flag and document them as not-yet-correct.

### Warnings

**W1. `current_uid()` via `metadata(home).uid()` is fragile.** `write.rs:43-55` reads the *owner* of `$HOME` to derive the running uid. On a shared dev box, sudo'd shell, or container with mounted home, returns the wrong uid and `find_tool_outputs` silently misses everything. Use `nix::unistd::getuid()` or `libc::getuid()` directly.

**W2. `rebuild_from_manifests` swap window leaves the index missing.** `index/mod.rs:194-206`: `remove_dir_all` then `rename`. If `rename` fails after the remove succeeds, the user has no index AND lost the previous one. Mitigation: rename old `by-project/` to `by-project.old`, rename staging into place, then rm old.

**W3. `--include-agents` without `--include subagents` is a silent no-op.** Both flags are genuinely orthogonal in the writer, but when `--clean --include-agents` is set without `--include subagents`, `find_agent_sessions` returns empty and the `--include-agents` becomes invisible. Either error at parse time or document the requirement in help text.

**W4. `find_history_slice_filters_by_window` doesn't actually test the function.** `sources.rs:317-354` — the test re-implements the inner loop against an inline string array. It exercises the test author's understanding of the algorithm, not the production code. Factor the line-filter logic into a pure helper and call it from both production and test.

**W5. `cwd_encoded` semantics are conflated with `project_path`.** `write.rs:381-388` passes `project_path.as_deref()` as `cwd_encoded` to `find_mcp_logs`. They happen to be the same string today, but a future read-side that decodes `project_path` to a real path would break this implicit contract. Rename the local var or wrap in a newtype.

### Suggestions

**S1.** `BulkSummary` defines `archived_count`/`skipped_count`/`archive_paths`; `ArchiveResult` defines the same three. Collapse — `BulkSummary` could just BE `ArchiveResult`, eliminating the field-by-field copy at `mod.rs:122-124`.

**S2.** `index/mod.rs:218`: when multiple archives have different absolute paths but the same basename slug, `ProjectEntry.absolute_path` arbitrarily picks the first. Stash the full set on `ProjectEntry` (e.g. `all_absolute_paths: Vec<PathBuf>`) so PR 3 doesn't have to re-walk for ambiguity detection.

**S3.** `IncludeSet::parse` returns `Result<Self>` but the body never returns `Err`. Either drop the `Result` or error on something (e.g., reject `none,all`).

**S4.** `eprintln!("warning: ...")` everywhere — `tracing::warn!` would let users silence/structure these. Not blocking.

**S5.** `paths.rs::extract_short_id` returns the last `-`-segment regardless of shape. `extract_short_id(".bashrc")` → `".bashrc"`. Add length/hex-char validation or document the precondition.

### Nitpicks

**N1.** `sources.rs:52-54` — `_session_modified` documented as "retained for future window-tightening." If PR 3 doesn't use it, drop it; if it does, drop the underscore.

**N2.** `index/mod.rs:227-231`: `eprintln!("Rebuilt by-project index: ...")` fires on every archive run. Gate behind verbosity or move to `tracing::debug!`.

**N3.** `write.rs:585`: `archive_session` returns `Result<Option<PathBuf>>` with documented "always `Some` today." Either tighten to `Result<PathBuf>` or actually return `None` on a no-op.

**N4.** Help text on `--include` (cli.rs:1194) lists `tool-output` but the underscore alias `tool_output` is undocumented despite being supported. Pick one and document.

### What's good

The split between `mod.rs` / `write.rs` / `sources.rs` / `paths.rs` / `include.rs` is clean — no logic duplication between `save_session` and `run`. `IncludeSet::parse` test coverage is thorough (12 cases). `skip_serializing_if` guards on new manifest fields are present and the `Manifest` round-trip test in `codex/mod.rs:194-197` is doing real work — much stronger than the archive-level one. The relative-symlink choice is correctly implemented and `make_symlink` is `cfg(unix)`-gated. Best-effort post-write index rebuild is wired exactly as specified.

Fix C1 and C2, then the warnings. The rest is acceptable.

coryzibell:

**Verdict: clean nod. Ready to merge.**

All 16 findings verified across the 6 fix-commits:

- **C1** — Goldens at `tests/fixtures/manifest-golden/{single-full,single-clean}.json` are real serialized JSON with sentinel substitution for volatile fields only. Comparison logic at `archive/mod.rs:303-348` does both structural `Value`-equality AND pretty-print byte-identity — a field reorder, rename, indentation drift, or serializer-setting change all fail. `MX_UPDATE_GOLDENS=1` regen path is operator-gated. Two `--all` cells documented as TODO at `mod.rs:444-457`.
- **C2** — `derive_session_window` (`sources.rs:65-125`) walks JSONL `timestamp` fields (string + epoch-seconds + epoch-millis), defensively orders the window if shuffled, falls back to `mtime`/`created()` with stderr warning on empty/garbage. Manifest `session_start`/`session_end` consume the same window. Consistency holds.
- **W1** — `current_uid()` uses `unsafe extern "C" { fn getuid() -> u32; }` gated on `#[cfg(unix)]`. Unsafe block narrowly scopes the single syscall. SAFETY comment correct.
- **W2** — Three-step swap with rollback at `index/mod.rs:215-269`: pre-clean stale `.old`, rename aside, rename in, cleanup or restore. Tests cover happy-path-no-leak and stale-`.old` sabotage. Deviation acceptable — the load-bearing invariant ("prior tree survives errored rebuild") is proven.
- **W3** — Validation at `handlers/mod.rs:99-105` errors clearly. Silent no-op is gone.
- **W4** — `filter_history_lines` pure helper at `sources.rs:294`, called by both production (`find_history_slice`) and test. Real code now exercised.
- **W5** — Local renamed to `cwd_encoded` at `write.rs:290-295` with identity-bridge comment for a future decoder.
- **S1** — `BulkSummary` collapsed; comment at `archive/mod.rs:116`.
- **S2** — `ProjectEntry::absolute_paths: Vec<PathBuf>` + `first_absolute_path()` helper. Ambiguous-basename test added.
- **S3** — `IncludeSet::parse` rejects mid-list empty tokens and `all+none` combo.
- **S4** — Skipped per brief; `tracing` confirmed not a dep.
- **S5** — `extract_short_id` validates non-empty + ASCII alnum trailing segment with graceful fallback.
- **N1** — `_session_modified` parameter removed.
- **N2** — Index-rebuild log gated on `MX_VERBOSE`.
- **N3** — `archive_session` returns `Result<PathBuf>`; no `Option` wrapper.
- **N4** — `tool_output` underscore alias removed; canonical `tool-output` only.

**Cross-cutting:** Diff stat only grows the codex archive tree + `cli.rs`/`handlers/mod.rs` deltas matching the changes. No collateral file growth. Status-quo manifests still round-trip; the new C1 goldens are stricter than the prior byte-identity claim.

PR #270, opened 2026-04-30
feat(codex): mx codex export + detection warning + index reads (PR 3/5 of #254)

coryzibell:
**Part of:** #254 (PR 5 closes the umbrella)

**Architecture context:** https://github.com/coryzibell/mx/issues/254#issuecomment-4349828527

**Prior PRs:** #267 (foundation), #268 (archive skeleton)

## Summary

Lands `mx codex export`, the read side of the codex story. The new subcommand reads exclusively from the codex (never `~/.claude/`), filters by session/project/date, and emits markdown or structured JSON. It implements `ProjectIndex::lookup` and `is_stale` -- the two methods stubbed since PR 1. A detection warning fires when unarchived sessions/tool-outputs are found in live sources, and `--archive-first` is offered as a one-shot capture-and-emit convenience. Default `--include` is `subagents` only -- clean human conversation by default.

## Files affected

- **New module:** `src/codex/export/{mod,filter,detect,read,include}.rs` + `src/codex/export/format/{mod,markdown,json}.rs`
- **Index implementation:** `src/codex/index/mod.rs` (`lookup` and `is_stale` no longer return `NotImplemented`; new `IndexError::NotFound` variant)
- **CLI wiring:** `src/cli.rs` (`CodexCommands::Export`), `src/handlers/mod.rs` (`handle_codex_export`)
- 8 new files, 12 files modified total; +2804 net LOC

## CLI surface

```
mx codex export
  [--session <uuid-or-short>]
  [--project <abs-path|slug|basename>]
  [--date <YYYY-MM-DD|range|YYYY-MM>]
  [--format markdown|json|both]
  [--include <comma-list>]   (default: "subagents")
  [--archive-first]
  [-o, --output <path>]
```

`--session`, `--project`, and `--date` are mutually exclusive; if none are given, "latest" by `session_start`.

## Detection warning

Runs at the start of every `export` invocation. Scans `~/.claude/projects/` (overridable via `MX_CLAUDE_PROJECTS_DIR` for testing) and `/tmp/claude-<uid>/.../tasks/`. Prints a stderr note with count and sample UUIDs (capped at 5). `--archive-first` short-circuits the warning by running `archive::run(All)`, re-detecting, and proceeding to export.

## Test plan

- [x] `cargo build` clean
- [x] `cargo test` 630 / 0 / 3 (+65 from the 565 baseline post-PR-#268)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
[...]

PR #271, opened 2026-04-30
refactor(session): mx session export becomes deprecation alias (PR 4/5 of #254)

coryzibell:
**Part of:** #254

**Architecture context:** [Compatibility section in #254 design comment](https://github.com/coryzibell/mx/issues/254#issuecomment-4349828527)

**Prior PRs (all merged):** #267 (foundation), #268 (archive skeleton), #270 (codex export)

## Summary

Shrinks `src/session.rs` from a 216-line standalone implementation to a ~156-line deprecation alias that translates `mx session export`'s flag surface into an `ExportRequest` and calls into `export::run`. Prints a stderr deprecation note pointing users at `mx codex export` and surfacing the new richer default (sub-agent transcripts now included by default — the old command silently dropped them). Forces `archive_first: true` so users without an existing codex still get a transcript (alias mirrors live-read semantics from the user's perspective).

## Files affected

- `src/session.rs` (216 → 156 non-test LOC; +237 test LOC — legacy module had no tests at all before this PR)
- `src/cli.rs` (+13/-3, help text only)

## Behavior changes (documented in the deprecation note)

- **Sub-agent transcripts now included by default.** The old `mx session export` silently dropped them; the new alias inherits `mx codex export`'s default `--include subagents` behavior. Users can pass `--include none` to suppress.
- **Output is sourced via the codex.** Behind the scenes, the alias forces `--archive-first`, so a user who never ran `mx codex archive` still gets their transcript (the alias archives transparently, then exports). User-perceived behavior matches the old live-read flow.

**CLI surface preserved byte-for-byte:** `mx session export [PATH] [-o OUTPUT]` accepts the same args as today. Existing user scripts continue to work.

## Test plan

- `cargo build` clean
- `cargo test`: 652 passed / 0 failed / 3 ignored (+11 from the 641 baseline post-PR-#270; 10 new session tests, 1 net incidental)
- 10 new tests assert: deprecation warning on stderr; empty positional → `Selector::Latest`; specific JSONL path positional → `Selector::Session(uuid)`; `archive_first: true` is set; output path is propagated; alias produces same shape as `mx codex export --session <uuid>`; flag mapping is byte-for-byte

## One subtlety preserved

`find_most_recent_session` survives in `src/session.rs` (was migrated-out of the alias path but `codex::archive::resolve_session_path` still consumes it for the "archive the live session right now" code path). Documented in rustdoc with a note that it migrates when `src/session.rs` deletes (PR 4-followup, one release out).

## Notable

The known `surreal_db::tests::test_open_with_path` flake did not fire (8th confirmed shift; pre-existing; not in scope here).

## What this PR does NOT do

- Does NOT delete `src/session.rs` entirely — that's a future PR after one release of warning
- Does NOT add `--backfill` or vault-detected warning (PR 5)
- Does NOT touch archive, codex, or any other module beyond the session alias and CLI help text

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1159,12 +1159,22 @@ pub enum CommentCommands {
 
 #[derive(Subcommand)]
 pub enum SessionCommands {
-    /// Export session to markdown
+    /// Export a Claude session as markdown.
+    ///
+    /// DEPRECATED: use `mx codex export` instead. This subcommand is now
+    /// a thin alias that forwards into the codex export pipeline and
+    /// will be removed in a future release. The new command supports
+    /// filtering by --session, --project, --date, multiple output
+    /// formats (markdown / json / both), and inlines sub-agent
+    /// transcripts by default.
     Export {
-        /// Path to session JSONL file (defaults to most recent non-agent session)
+        /// Path to session JSONL file (defaults to most recent non-agent session).
+        ///
+        /// DEPRECATED: when set, the file stem (the session UUID) is
+        /// extracted and routed to `mx codex export --session <uuid>`.
         path: Option<String>,
 
-        /// Output file (defaults to stdout)
+        /// Output file (defaults to stdout).
         #[arg(short, long)]
         output: Option<String>,
     },
--- a/src/session.rs
+++ b/src/session.rs
@@ -16,6 +16,7 @@
 
 use anyhow::{Context, Result};
 use std::fs;
+use std::io::{self, Write};
 use std::path::PathBuf;
 
 /// Stderr-printed deprecation notice for `mx session export`. Kept as a
@@ -85,6 +86,19 @@ pub(crate) fn build_alias_request(
     })
 }
 
+/// Write the deprecation notice to the given writer, followed by a
+/// trailing newline. Extracted as a generic helper so tests can inject
+/// a `Vec<u8>` and assert routing without wrestling with process-level
+/// stderr capture (no `gag` dev-dep needed). Production wires this to
+/// `io::stderr()`.
+///
+/// Load-bearing: the notice MUST go to stderr, not stdout — pipelines
+/// like `mx session export | pandoc` would otherwise be corrupted by
+/// the warning text appearing in the document body.
+pub(crate) fn print_deprecation_notice<W: Write>(w: &mut W) -> io::Result<()> {
+    writeln!(w, "{}", DEPRECATION_NOTICE)
+}
+
 /// Entry point invoked by `handle_session(SessionCommands::Export {..})`.
 ///
 /// Prints the deprecation notice to stderr, builds the request, and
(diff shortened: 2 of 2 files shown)


coryzibell:

### Critical
None. The shape is correct. The alias preserves `mx session export [PATH] [-o]`, routes positional `PATH` to `Selector::Session(file_stem)`, falls back to `Selector::Latest`, forces `archive_first: true`, propagates `output`, and prints `DEPRECATION_NOTICE` on stderr before any work.

### Warnings

**W1. No test asserts the deprecation notice actually reaches stderr.** `deprecation_notice_mentions_new_command` (lines 165-177) only checks the constant's text. `export_session` is the function that does `eprintln!`, and nothing exercises that path. If a refactor swaps `eprintln!` for `println!` tomorrow, every existing test still passes, and piped consumers of `mx session export | foo` start choking on the warning in their stdout. Add a test that calls `export_session` and asserts the notice landed on stderr — or at minimum a test that grep's the source for `eprintln!(.*DEPRECATION_NOTICE)`. The brief specifically called out "printed to stderr (NOT stdout — would corrupt piped output)" as load-bearing.

**W2. The end-to-end equivalence test flips `archive_first` off (line 338).** Honest about why — re-archiving a tempdir with no live sources — but this means no test exercises `archive_first: true` through `codex::export::run`. Unit-level routing assertion on line 209 is fine for the alias claim; the integration story is "I trust `codex::export::run` honors the flag" (which PR #270 tests cover). Acceptable as written, but document this gap in either a code comment or the PR description.

### Suggestions

**S1.** `find_most_recent_session` is `pub`, not `pub(crate)` (line 108). The module doc-comment says it's kept for `codex::archive::resolve_session_path` only. If in-crate is the only consumer, tighten visibility.

**S2.** Path validation is loose (lines 64-76). `mx session export README.md` would route to `Selector::Session(SessionRef("README"))` and fail downstream with codex-flavored "session not found." Legacy command also failed on bad input, so no worse — but a stricter check (`stem` looks UUID-shaped, or the path ends in `.jsonl`) would yield a better error.

**S3.** The `Selector::Latest` vs `find_most_recent_session` semantic drift the brief flagged. With `archive_first: true` set, the codex archives `--all` unarchived sessions, then `Selector::Latest` picks by `session_start`. For users with a single live session this matches legacy mtime-of-JSONL behavior. For users with several unarchived sessions it might not. Worth a single sentence in `DEPRECATION_NOTICE` ("selects the most recent session by start time, not file mtime") if you want to be thorough.

Fix W1 — add a stderr-capture test for the deprecation notice. W2 and the suggestions are author's discretion. The rest is acceptable.

coryzibell:

**Verdict: clean nod. Ready to merge.**

All 5 findings verified across the 3 fix-commits:

- **W1** — `print_deprecation_notice<W: Write>` at `session.rs:157`. Production call at `:171` is `print_deprecation_notice(&mut io::stderr())`. Injection test writes to `Vec<u8>` and asserts the full notice. Source-text guard strips the test module before grepping (avoids matching its own assertion string) and negates `println!("{}", DEPRECATION_NOTICE)`. A swap to stdout would fail both tests.
- **W2** — Equivalence-test comment at `:506-513` explains the `archive_first = false` flip, names the determinism reasoning, points to `alias_forces_archive_first` and PR #270 coverage.
- **S1** — `find_most_recent_session` is `pub(crate)` at `:189` with doc-comment justifying the choice and flagging the future migration. Sole external caller is `src/codex/archive/mod.rs:172`. Zero `tests/` references.
- **S2** — `is_uuid_shaped` at `:119` handles full 36-char dashed UUID and 8-char hex short form. Rejects `README.md` with the expected-shape message and `mx codex export --session` hint. `c3744b8d` (bare short UUID without extension) works. `zzzzzzzz` correctly rejected via `is_ascii_hexdigit`.
- **S3** — `DEPRECATION_NOTICE` (`:32-34`) adds "session start time, not JSONL file mtime" clarification with single-session caveat.

**Cross-cutting:** Alias contract intact — notice fires before request build, `archive_first: true` retained, selector resolution unchanged. 15/15 session tests pass. No regressions.

PR #272, opened 2026-04-30
feat(codex): vault backfill + presence warning, closes #254 (PR 5/5)

### LINKED ISSUE #254: combine mx codex archive and mx session export -- overlapping ingestion of Claude sessions

coryzibell:
**Closes #254.** This is the final PR of the 5-PR codex unification series. The umbrella issue tracks the full architecture: https://github.com/coryzibell/mx/issues/254#issuecomment-4349828527

**Prior PRs:** #267 (foundation: paths/index/manifest v5), #268 (archive skeleton + walkers + index writes), #270 (codex export + detection + index reads), #271 (session export deprecation alias)

## Summary

Adds `mx codex save --backfill [<vault-path>]` to ingest the historical session data still living in `~/.wonka/vault/archives/` (26 vault snapshots, ~28M of session data on this machine). Default vault path is `paths::wonka_vault_archives_dir()` if no argument given. Adds a vault-presence warning that fires once per process at the top of every `mx codex *` handler when `~/.wonka/vault/archives/` contains session snapshots — suppressed during `--backfill` invocations to avoid double-nagging users in the act of backfilling. Per-session failures during backfill are non-fatal; aggregated into the `BackfillReport` and surfaced at completion.

## Files affected

- **New module:** `src/codex/archive/backfill.rs` — vault walker, `BackfillReport` struct, 4 tests (idempotency, missing-vault error, empty-vault zero-report, normal walk)
- **New module:** `src/codex/notices.rs` — `warn_if_vault_present(suppress)` with `OnceLock` once-per-process guard, 4 tests (missing/empty/junk-only/has-snapshot)
- **CLI wiring:** `src/cli.rs` (new `--backfill` flag with `conflicts_with_all` against `--all` and the positional `[PATH]`; optional value via `num_args = 0..=1`)
- **Handler:** `src/handlers/mod.rs` (`handle_codex` calls `warn_if_vault_present` at top, computes suppress via `matches!` on the backfill flag)
- 6 files modified total; +621/-2 net LOC

## CLI surface addition

```
mx codex save --backfill [<vault-path>]
```

Optional positional argument; defaults to `~/.wonka/vault/archives/`. Mutually exclusive with `--all` and the positional `[PATH]`. The other capture flags (`--include`, `--clean`) still apply.

## Naming note

The user-facing CLI subcommand is `mx codex save` (existing); the architecture proposal called it `mx codex archive` throughout for clarity. The internal entry point is `archive::run`, but the CLI verb is `save`. A future PR could rename `save` → `archive` for full consistency with the architecture; out of scope for this series.

## Idempotence

`--backfill` is idempotent — running twice on the same vault produces the same final codex state. `archived_ids` dedup logic was rebuilt locally inside `run_backfill` (it only existed in the `save_all_sessions` `--all` path before; `Single` does not dedup natively). Verified by `backfill_is_idempotent` test.

## Vault warning behavior

- Vault present + has session-* snapshots → fires once per process
- Vault absent → silent
- Vault present but empty (no `session-*` children) → silent
- `--backfill` invocation → suppressed

## Test plan

- [x] `cargo build` clean
- [x] `cargo test` 664 passed / 0 failed / 3 ignored (+8 from 656 baseline post-PR-#271)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [x] The known `surreal_db::tests::test_open_with_path` flake fired (9th confirmed shift; pre-existing; not in scope here)

## Out of scope (reviewer expectations)

- Does NOT delete the vault directory or any of its contents (user's call after verifying backfill)
- Does NOT touch `~/.wonka/bin/activate` (separate post-series factory commit will replace the vault-write block with `mx codex save` integration)
- Does NOT delete `src/session.rs` (future post-series PR after one release of deprecation warning)
- Does NOT capture tool-output or MCP sidecars from vault snapshots — those were never archived to the vault (ephemeral by the time backfill runs)

**Series complete after merge.** Five PRs, no broken intermediate states, status-quo behavior preserved at every step until users explicitly opt in via `--include` or `mx codex export`. The factory's `~/.wonka/bin/activate` change becomes a follow-on once this lands.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1185,10 +1185,11 @@ pub enum CodexCommands {
     /// Archive current session to permanent storage
     Save {
         /// Path to session JSONL file (defaults to most recent non-agent session)
+        #[arg(conflicts_with_all = ["all", "backfill"])]
         path: Option<String>,
 
         /// Archive all unarchived sessions
-        #[arg(long)]
+        #[arg(long, conflicts_with_all = ["backfill"])]
         all: bool,
 
         /// Save only conversation.md + manifest.json + images (no JSONL, no agent files)
@@ -1216,6 +1217,26 @@ pub enum CodexCommands {
         /// `conversation.md` rendering when `--clean` is set.
         #[arg(long, default_value = "subagents")]
         include: String,
+
+        /// Ingest the legacy wonka vault snapshots into the codex.
+        ///
+        /// Walks every `session-*` snapshot under the supplied path
+        /// (defaults to `~/.wonka/vault/archives/`) and feeds each
+        /// session JSONL through the regular archive pipeline.
+        /// Idempotent: re-running on the same vault produces the same
+        /// codex state.
+        ///
+        /// Mutually exclusive with `--all` and the positional `[PATH]`.
+        /// `--include` and `--clean` still apply (they govern what the
+        /// per-session pipeline captures).
+        #[arg(
+            long,
+            value_name = "VAULT_PATH",
+            num_args = 0..=1,
+            default_missing_value = "",
+            conflicts_with_all = ["all", "path"],
+        )]
+        backfill: Option<String>,
     },
 
     /// Export an archived session as Markdown or structured JSON.
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1277,7 +1277,7 @@ pub enum CodexCommands {
         #[arg(long, default_value = "subagents")]
         include: String,
 
-        /// Run `mx codex archive --all` before exporting and skip the
+        /// Run `mx codex save --all` before exporting and skip the
         /// unarchived-data warning. Useful when you know live
         /// `~/.claude/projects/` data hasn't been ingested yet.
         #[arg(long)]
--- a/src/codex/archive/backfill.rs
+++ b/src/codex/archive/backfill.rs
@@ -339,7 +339,14 @@ mod tests {
     }
 
     /// Drive a backfill against an isolated codex dir + fake vault.
-    fn drive_backfill(n_snaps: usize, sess_per_snap: usize) -> (BackfillReport, PathBuf) {
+    /// Returns the `TempDir` so the caller's scope keeps it alive
+    /// (dropping it cleans up the codex + vault on test exit). The
+    /// previous version called `tmp.keep()` which leaked the directory
+    /// onto disk on every test run — fixed per S2 in PR 272 review.
+    fn drive_backfill(
+        n_snaps: usize,
+        sess_per_snap: usize,
+    ) -> (BackfillReport, PathBuf, tempfile::TempDir) {
         let _guard = ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner());
         let tmp = tempfile::tempdir().unwrap();
         let codex_dir = tmp.path().join("codex");
@@ -362,11 +369,7 @@ mod tests {
             }
         }
 
-        // Hold tmp dir for the caller to inspect. Caller's responsibility
-        // to keep it alive long enough; here we leak it via PathBuf and
-        // accept the temp-dir orphan for the duration of the test.
-        let leaked = tmp.keep();
-        (report, leaked.join("codex"))
+        (report, codex_dir, tmp)
     }
 
     /// Add a "corrupt" session JSONL to an existing vault snapshot. The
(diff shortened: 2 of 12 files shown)


coryzibell:

### Critical

**C1. Vault-warning text directs operator to a non-existent subcommand.** `src/codex/notices.rs:38-39` emits:

> `Run \`mx codex archive --backfill\` to ingest, then remove the vault directory.`

The `--backfill` flag is wired onto `mx codex save`, not `mx codex archive`. The PR body acknowledges this naming gap explicitly. An operator who follows the warning verbatim hits "no such subcommand." The warning is the entire user-facing affordance for the backfill feature; it must be runnable as written. Fix: change the message to `mx codex save --backfill` and update the doc comment on lines 13-15 that says the same thing.

### Warnings

**W1. `collect_archived_ids` swallows all read errors silently via `unwrap_or_default()` (line 95).** A permission-denied or corrupted manifest produces zero dedup, which means re-running backfill after a partial failure could double-archive. Suggest at least an `eprintln!` when the precompute fails wholesale.

**W2. `session_id_from_path` derives the id from the filename stem.** Works for the documented vault layout, but if `archive::run(Single)` ever computes a different canonical id (e.g., from JSONL `session_id` field), the dedup key drifts and idempotence quietly breaks. Worth a comment in `archive::run` pointing back here so the next refactorer doesn't change the convention silently.

### Suggestions

**S1.** The "archive vs save" naming inconsistency the PR body flags is now baked into operator-visible warning copy and module docs. File the rename follow-up as part of closing #254 so it doesn't drift. The fix in C1 is a stopgap, not the resolution.

**S2.** `backfill.rs:342` uses `tmp.keep()` to leak the tempdir for the test's lifetime. Functionally fine for `cargo test` but leaves stray dirs under `$TMPDIR` across runs.

**S3.** No test exercises a per-session failure (corrupt JSONL feeding `archive::run`). The non-fatal-error claim is only structural, not behavioral. One test that drops a malformed `.jsonl` into the fake vault and asserts `report.errors.len() == 1` while the other sessions still archive would close the loop.

### Notes

- **Idempotence: verified.** Pre-walk dedup + cross-snapshot `seen` set is correct. The two-pass test is real.
- **OnceLock:** process-local static, correct scope. Suppress takes precedence before the lock check.
- **CLI wiring:** `num_args = 0..=1` + `default_missing_value = ""` + handler fallback is the right shape. `conflicts_with_all` covers both collisions.
- **Series closer is coherent:** live → save → codex → export → user.

Fix C1. The rest is acceptable but per loop convention, address W1/W2 and S1/S2/S3 in this PR — it's the closer.

coryzibell:
S1: filed follow-up issue: https://github.com/coryzibell/mx/issues/273

coryzibell:

**Verdict: clean nod. Ready to merge. Series complete.**

All 6 findings verified across 5 fix-commits + 1 follow-up issue:

- **C1** — Subcommand sweep is literal-string-only. Internal `archive::run`, `archive::mod`, `collect_archived_ids`, `session_id_from_path` all untouched. Regression test in `notices.rs:117-153` slices the test module out before grepping the function body — avoids the self-defeating "test contains the forbidden string" trap.
- **W1** — `collect_archived_ids` failure surfaces on BOTH stderr AND `report.errors` with `.context()` wrapping. Test forces `NotADirectory` via codex-as-file path.
- **W2** — Reciprocal NOTE comments in `archive::run` Single arm and `session_id_from_path`. Both name the other by module path. Dedup-key invariant explicit.
- **S1** — Filed as #273; out-of-band per loopa convention.
- **S2** — `drive_backfill` returns `(BackfillReport, PathBuf, TempDir)`. No `.keep()`. Caller scope owns lifetime.
- **S3** — 3 valid sessions + 1 chmod-000 corrupt; asserts `4 found / 3 archived / 1 error / Ok return`.

**Cross-cutting:** No regressions. Diff coherence holds — the 9-file sweep is all string-substitution, not logic. The flake fired (10th shift; known noise).

Ship the closer.

PR #275, opened 2026-04-30
fix(ci): cfg(unix) gates + graceful missing claude projects dir (CI hotfix)

coryzibell:
## HOTFIX

CI on `main` has been failing since PR #267 merged. This PR stops the bleeding.

Each individual PR (#267-272) passed PR-level CI, but post-merge CI on main exposed:

1. **Windows compile failures** — un-gated `Permissions::set_mode()` in PR 5's S3 fix test
2. **macOS+Ubuntu test failure** — `export_archive_first_skips_warning` panicked because `archive::run(All)` errored on missing `~/.claude/projects/` (CI machines never ran Claude)

**Failing CI run for reference:** https://github.com/coryzibell/mx/actions/runs/25184704339 (commit `1343d84`)

### Commit `a6a97ca` — `cfg(unix)` gates

`#[cfg(unix)]` gate on the `chmod 000` test + helper in `backfill.rs`. The third unix call site (`make_symlink` in `index/mod.rs:825`) was already correctly gated.

Verified: `rg "set_mode|PermissionsExt|std::os::unix" src/` returns only sites inside `cfg(unix)` blocks.

### Commit `4fa85e1` — graceful missing claude projects dir

Twofold fix:

- `save_all_sessions` (`archive/write.rs:548`) replaced its hard `anyhow::bail!` on missing projects dir with stderr warning + `ArchiveResult::default()`. Fresh users (and CI) now get a graceful no-op instead of a crash.
- `paths::claude_projects_dir()` was IGNORING the `MX_CLAUDE_PROJECTS_DIR` env override — meaning every test that "set" the override was actually walking the developer's real `~/.claude/projects/`. The override is now respected. PR 3's `MX_CLAUDE_PROJECTS_DIR` design always intended this; the wiring was incomplete. **This is the hidden bug behind the CI failure.**
- Two `paths.rs` tests (`claude_paths_under_home`, `claude_subagents_dir_layout`) marked `#[serial]` with env clear/restore to avoid racing with codex tests that set the override.

## Files affected

- `src/codex/archive/backfill.rs` — `cfg(unix)` gates
- `src/codex/archive/write.rs` — graceful no-op on missing projects dir
- `src/paths.rs` — env override wired in; serial-test guards on existing tests

3 files, +67/-1 net.

## Test plan

- [x] `cargo build` clean (Linux verified; Windows compile previously failed at the unix gate, expected to pass now)
[...]

PR #279, opened 2026-05-01
fix(test): clear MX_CLAUDE_PROJECTS_DIR in dir_missing test

### LINKED ISSUE #276: Test hardening: clear MX_CLAUDE_PROJECTS_DIR in find_most_recent_session_errors_when_dir_missing
## What

`find_most_recent_session_errors_when_dir_missing` (`src/session.rs:545`) sets `HOME` to a nonexistent path but does not also clear `MX_CLAUDE_PROJECTS_DIR`.

After PR #275 wired the env override into `paths::claude_projects_dir()`, a leaked `MX_CLAUDE_PROJECTS_DIR` from a prior test in the same process could swing this test's assertion. Today the test is safe in practice — it carries `#[serial]` and uses `ENV_LOCK` — but defense-in-depth says clear the override explicitly alongside `HOME`.

## Where

\`src/session.rs:545\` — \`find_most_recent_session_errors_when_dir_missing\`

## Suggested fix

Capture and clear \`MX_CLAUDE_PROJECTS_DIR\` at test start; restore at end (mirror the \`prev_proj\` pattern used by the codex tests). One-line addition.

## Severity

Low. Pre-existing latent. Surfaced by Verdictia during PR #275 review; she explicitly flagged it as non-blocking.

## Reference

PR #275 (CI hotfix): https://github.com/coryzibell/mx/pull/275

coryzibell:
Closes #276

## Summary

Clears `MX_CLAUDE_PROJECTS_DIR` at test start and restores it at end, mirroring the existing capture/clear/restore pattern for `HOME` and `MX_HOME`. Defense-in-depth against env leakage after PR #275 wired the override into `paths::claude_projects_dir()`.

## Files affected

- `src/session.rs`

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -549,10 +549,12 @@ mod tests {
 
         let prev_home = std::env::var("HOME").ok();
         let prev_mx_home = std::env::var("MX_HOME").ok();
+        let prev_proj = std::env::var("MX_CLAUDE_PROJECTS_DIR").ok();
         // SAFETY: serial + lock.
         unsafe {
             std::env::set_var("HOME", &nonexistent);
             std::env::remove_var("MX_HOME");
+            std::env::remove_var("MX_CLAUDE_PROJECTS_DIR");
         }
         let result = find_most_recent_session();
         unsafe {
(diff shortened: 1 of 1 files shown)


coryzibell:
## Code Review

**W1 — Inconsistent restore pattern for `MX_HOME` (pre-existing)**
The new `MX_CLAUDE_PROJECTS_DIR` restore correctly uses `match` with both `Some`/`None` arms. The existing `MX_HOME` restore only handles `Some` — if `MX_HOME` wasn't set before the test, it could leak. Pre-existing; not a blocker. Worth fixing as follow-up.

**S1 — Panic-safety (pre-existing)**
If `find_most_recent_session()` panics, restore logic doesn't run. Low probability on an error-path test, but a scope guard would be bulletproof. Pre-existing pattern.

**N1 — Variable name `prev_proj`**
Consistent with usage in the other test at line 482. Acceptable.

**Verdict:** Ship. The change is correct and actually improves on the pattern it mirrors.

PR #280, opened 2026-05-01
feat(commit): add --dry-run flag

### LINKED ISSUE #201: mx commit: add --dry-run flag
## Summary

`mx commit` has no way to preview what it would commit without actually committing and pushing. Add a `--dry-run` flag that prints the would-be output and side effects without mutating git state.

## Motivation

Two reasons:

**1. Review safety.** During the review of #200, a reviewer empirically testing the clap constraint graph on the new `--show-encoded` flag accidentally ran `./target/debug/mx commit --show-encoded --all --push "msg"` in the mx working tree. `--all` swept in untracked files; `--push` sent the result to origin as a stray commit on the PR branch. Caught and rolled back within seconds via `git reset --hard` + `git push --force-with-lease`, but avoidable. A `--dry-run` flag would make this class of near-miss impossible: reviewers could verify CLI constraint graphs empirically without risking state mutation.

**2. User sanity check.** Users who want to preview the encoded title/body/footer that `mx commit` will actually write currently have no clean way to do so. This is especially relevant after #200 changes the default stdout from verbose to footer-only: seeing the full encoded form before committing is now an opt-in (`--show-encoded`), and there's no way to see it *without* committing.

## Design sketch

Not binding, just a starting point:

- `mx commit --dry-run "msg"` — print what would be committed, skip `git_commit` and `git_push`
- Compatible with `--show-encoded`, `-a`, `-p`, `--encode-only` (all no-ops under `--dry-run`)
- Exit 0 if dry-run succeeds (encoding completes, git status is commit-ready)
- Exit nonzero if the real commit would have failed (no staged changes, encoding error, etc.)
- Output prefixed with `[dry-run]` or similar so it's never mistaken for a real commit log

## Context

Surfaced as an S3 (suggestion-grade nit) during Verdictia Stern's review of #200. Not blocking that PR — filing as a follow-up.

coryzibell:
Closes #201

## Summary

Adds a `--dry-run` flag to `mx commit` that runs all encoding and validation logic but skips the actual `git commit` and `git push`. Output is prefixed with `[dry-run]`. Exits 0 on success, nonzero if the commit would have failed. Conflicts with `--encode-only`. Compatible with `--show-encoded`, `-a`, and `-p`.

## Motivation

Prevents accidental commits when testing CLI behavior — a near-miss during the PR #200 review highlighted the need for a safe way to exercise the full commit pipeline without mutating git state.

## Files changed

- `src/cli.rs` — adds `--dry-run` flag definition and conflict with `--encode-only`
- `src/commit.rs` — dry-run gate around git commit/push, `[dry-run]` output prefix
- `src/main.rs` — threads the dry-run flag through to the commit handler

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -52,6 +52,12 @@ pub enum Commands {
         /// Default output is just the footer line and `Committed.`
         #[arg(long, conflicts_with = "encode_only", conflicts_with_all = ["title", "body"])]
         show_encoded: bool,
+
+        /// Preview what would be committed without mutating git state.
+        /// Runs all encoding/validation logic but skips git commit and push.
+        /// Output is prefixed with `[dry-run]`.
+        #[arg(long, conflicts_with = "encode_only", conflicts_with_all = ["title", "body"])]
+        dry_run: bool,
     },
 
     /// Pull request operations
--- a/src/commit.rs
+++ b/src/commit.rs
@@ -479,6 +479,19 @@ pub fn format_encoded_commit(encoded: &EncodedCommit, show_encoded: bool) -> Str
     out
 }
 
+/// Prefix every line of `output` with `[dry-run] `.
+///
+/// Useful for marking preview output so it is never mistaken for a real
+/// commit log. Extracted as a standalone helper so it can be unit-tested
+/// without git state.
+pub fn prefix_dry_run(output: &str) -> String {
+    output
+        .lines()
+        .map(|line| format!("[dry-run] {}", line))
+        .collect::<Vec<_>>()
+        .join("\n")
+}
+
 /// Perform the full upload commit.
 ///
 /// `show_encoded` controls stdout verbosity:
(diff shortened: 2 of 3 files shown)


coryzibell:
## Code Review

### C1 — `git add -A` runs under dry-run, mutating the index
`src/commit.rs:503-506`. `stage_all()` under `--dry-run` obliterates carefully staged subsets. A `--dry-run` flag that rewrites the index violates its contract. Fix: skip `stage_all()` when `dry_run` is true.

### W1 — `--dry-run` does not conflict with `--push`
Semantically odd but informational. The `[dry-run] Would push.` message tells the user what would happen. **Decision: keep compatibility** — `-p` under dry-run is preview-of-intent, not a promise. The alternative (forcing the user to remove `-p` to dry-run) adds friction without safety.

### W2 — No tests
Zero tests for the new path. At minimum: unit test for dry-run output formatting, integration test asserting no commit is created.

### W3 — `[dry-run]` prefix logic is inline
Should live in `format_encoded_commit` or a helper so it's testable without git state.

### S1 — Error messages not prefixed with `[dry-run]`
`bail!()` on "no staged changes" has no prefix. User can't tell if the error came from a real or dry run.

### N1 — Duplicated `conflicts_with_all`
`dry_run` copies `show_encoded`'s conflict pattern. Brittle but correct. Consider `ArgGroup` later.

**Verdict:** Fix C1. Address W2/W3. The rest is acceptable.

coryzibell:
## Re-review

All four prior findings resolved:
- **C1** — `stage_all()` now gated by `!dry_run`. Index untouched.
- **W2** — 7 tests added covering prefix logic and output formatting.
- **W3** — `prefix_dry_run()` extracted as standalone testable helper.
- **S1** — Error messages include `[dry-run]` prefix.

**New N1:** `--all` is silently skipped under dry-run — user gets no hint. Correct behavior, minor UX gap. Not blocking.

**Verdict:** Ship.

PR #281, opened 2026-05-01
feat(pr): auto-switch to target branch and cleanup after merge

### LINKED ISSUE #217: mx pr merge: auto-switch to target branch and delete local PR branch after merge
## Problem

After `mx pr merge <N>` (squash merge, which is default), the GitHub remote deletes the source branch. But the local clone is still checked out on that now-deleted source branch. The next `mx commit -p` tries to `git pull --rebase` against `origin/<source-branch>` and fails:

```
Error: git pull --rebase failed: Your configuration specifies to merge with the ref
'refs/heads/fix/zed-libwebrtc-pipewire-patch' from the remote, but no such ref was fetched.
```

This hangs the workflow — the user has to manually:
1. `git checkout main`
2. `git reset --hard origin/main`
3. Redo any edits they'd made on the branch in the meantime (if any)

## Desired behavior

After a successful `mx pr merge <N>`, mx should:

1. `git fetch origin` — pick up the merged commit on the target branch + prune the deleted source ref
2. Read the PR's target branch from the API response (usually `main`, but could be anything)
3. `git checkout <target>` — switch the user back to the PR's target branch
4. `git pull --ff-only` (or equivalent) — land the merged commit locally
5. `git branch -D <source>` — delete the local source branch (it's dead upstream, no reason to keep it)

## Edge cases

- **Uncommitted changes on source branch:** don't auto-switch. Warn and bail with a clean error telling the user to stash/commit first.
- **Source branch still has unpushed local commits:** same — warn and bail, don't delete their work.
- **Target branch isn't `main`:** read from PR metadata, don't hardcode.
- **User workflows that keep source branches around:** `--no-cleanup` flag to opt out of step 5 (keep the branch delete, just don't do the auto-switch).
- **Rebase merge / merge-commit (non-squash):** for rebase merges, same behavior makes sense. For merge-commit, source branch is retained on the remote by default — still worth switching the user to the target, but the branch-delete step should check if the remote ref still exists.

## Evidence

Hit this twice in the 2026-04-17 → 2026-04-18 session working on the luola repo — once after `mx pr merge 80` (the libwebrtc pipewire fix), needed manual `git checkout main && git reset --hard origin/main` to continue working. The commit message would otherwise have gone to a dead branch with a failed push.

## Out of scope

- PR creation UX (this is merge-time only)
- Migrating existing state (this is a forward-looking behavior change)

coryzibell:
Closes #217

## Summary

- After a successful `mx pr merge`, automatically fetches+prunes, switches to the PR's target branch (read from API, not hardcoded), fast-forward pulls, and deletes the local source branch. Prevents the footgun where `git pull --rebase` fails on a dead remote ref.
- Guards against uncommitted changes and unpushed local commits — both bail with clear warnings before any cleanup begins.
- Adds `--no-cleanup` flag to skip all post-merge cleanup.
- All cleanup steps are best-effort — merge already succeeded, so failures are warnings not errors.

## Files changed

- `src/cli.rs`
- `src/commit.rs`
- `src/handlers/mod.rs`

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -269,6 +269,10 @@ pub enum PrCommands {
         /// Use standard merge commit instead of squash (mutually exclusive with --rebase)
         #[arg(long, name = "merge", conflicts_with = "rebase")]
         merge_commit: bool,
+
+        /// Skip post-merge cleanup (don't switch to target branch or delete local source branch)
+        #[arg(long)]
+        no_cleanup: bool,
     },
 }
 
--- a/src/commit.rs
+++ b/src/commit.rs
@@ -679,6 +679,20 @@ fn post_merge_cleanup(source_branch: &str, target_branch: &str) {
         return;
     }
 
+    // Guard: cleanup is pointless when source and target are the same branch
+    if source_branch == target_branch {
+        return;
+    }
+
+    // Guard: check if source branch exists locally -- if not, skip the
+    // unpushed-commits guard and branch deletion (nothing to check or delete)
+    // but still do fetch+prune, checkout target, and ff-only pull.
+    let source_exists_locally = Command::new("git")
+        .args(["rev-parse", "--verify", source_branch])
+        .output()
+        .map(|o| o.status.success())
+        .unwrap_or(false);
+
     // Guard: check for uncommitted changes before switching
     match has_uncommitted_changes() {
         Ok(true) => {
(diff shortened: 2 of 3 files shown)


coryzibell:
## Code Review

### C1 — No local branch existence check
`has_unpushed_commits` and `git branch -D` assume the source branch exists locally. If the user merges from `main` for a branch they never checked out, the flow stumbles through error swallowing to the right answer — but it's fragile. Add a `git rev-parse --verify` guard early.

### C2 — `git branch -D` should be `git branch -d`
Force-delete bypasses git's merge-status safety check, undercutting the guards above. Use `-d` — if the branch is properly merged, it succeeds; if not, it correctly refuses.

### W1 — `has_unpushed_commits` fragility
Related to C1. The `Err(_)` comment says "proceed cautiously" but the actual failure mode may be "branch doesn't exist locally." Comments should match reality.

### W2 — Untracked files block cleanup
`git status --porcelain` includes `??` lines. A stray `notes.txt` triggers "uncommitted changes" and skips cleanup. Should filter to tracked changes only.

### W3 — No tests
170 lines of branching logic with zero tests. At minimum, extract guard decisions into testable pure functions.

### S1-S3 — Structural
Decouple switch-to-target from delete-source. Print a summary line. Move `source == target` guard earlier.

### N1-N2 — Minor
Unit return type limits caller visibility. Hardcoded `"origin"` remote.

**Verdict:** Fix C1 and C2. Address W2. The rest at discretion.

coryzibell:
## Re-review

All four findings resolved:
- **C1** — `git rev-parse --verify` guard added. Skip unpushed-commits check and deletion when source branch doesn't exist locally. Fetch/checkout/pull still proceed.
- **C2** — `git branch -d` (safe delete). Refuses unmerged branches.
- **W2** — `git diff --quiet HEAD` replaces `git status --porcelain`. Untracked files no longer block cleanup.
- **S3** — `source == target` guard moved to top, above all work.

No new findings. **Verdict:** Ship.

PR #291, opened 2026-05-05
fix: resolve chronic test_open_with_path flake

### LINKED ISSUE #274: surreal_db::tests::test_open_with_path: intermittent flake (10+ shift pattern)
## Symptom

`src/surreal_db/tests.rs::test_open_with_path` opens a SurrealKV database at a `tempdir()`-derived path, then asserts `db_path.exists()` and `db_path.is_dir()`. Intermittently, one of those assertions fails (most commonly `is_dir`), even though the test passes on re-run with no code changes.

## Evidence

10+ confirmed firings across recent shifts and PRs: #259, #261, #262, #265, #266, #267, #268, #270, #271, #272 (codex unification series and prior shift work). Always passes on re-run; never blocks merge; treated as known noise per the loopa convention. Prior factory note: memory entry `kn-a8ed7a73` (5-shift pattern at filing time, anchor `kn-0cb79a83`).

## Hypothesis

Most likely a filesystem-creation race inside the SurrealKV embedded backend. `open_embedded_async` (`src/surreal_db/connection.rs:246`) calls `std::fs::create_dir_all` for the *parent* only, then `Surreal::new::<SurrealKv>(path).await`. The test asserts on `db_path` itself — created by SurrealKV. Under parallel `cargo test` load, SurrealKV's directory creation may not be visible to the calling thread's `std::fs::metadata` call by the time `Surreal::new` resolves.

Secondary suspect: `MX_SURREAL_MODE` environment leakage between parallel tests — if another test sets `network` mode in-process, `open_embedded_async` is skipped and `db_path` is never created, so the assertion fails.

## Reproducibility

Intermittent. Re-running `cargo test` typically passes. No reliable repro known.

## Suggested investigation directions

1. Run with `--test-threads=1` over many iterations; if flake disappears, it confirms parallel-test interference (likely env var or `/tmp` contention).
2. Add stderr instrumentation inside the test on failure: print `db_path`, `path.exists()`, `path.parent().exists()`, `MX_SURREAL_MODE` value, and a directory listing of the tempdir.
3. After `Surreal::new` returns, explicitly `fs::create_dir_all(&db_path)` in `open_embedded_async` (or in the test) and re-check — if that resolves it, the SurrealKV constructor is racy w.r.t. visible directory state.

## Acceptance

Fix lands when the test runs reliably across 10+ consecutive `cargo test` invocations on a clean checkout (parallel and `--test-threads=1`).

## References

- Memory entry `kn-a8ed7a73` (prior factory notes, 5-shift pattern)
- Anchor `kn-0cb79a83` (3rd-sighting note)
- Test source: `src/surreal_db/tests.rs:17`
- Open path: `src/surreal_db/connection.rs:246` (`open_embedded_async`)

coryzibell:
Closes #274

## Summary

- Force embedded SurrealDB mode in `test_open_with_path` by using `SurrealDatabase::connect()` with `SurrealConfig::default()` instead of `SurrealDatabase::open()`

## Root Cause

`open()` reads `MX_SURREAL_MODE` from the environment. On hosts where the factory runs in production (`MX_SURREAL_MODE=network`), the test connected to the remote SurrealDB server instead of creating a local embedded database. The local path was never created, so `db_path.exists()` / `is_dir()` assertions failed.

The flake was "intermittent" because it only fired when tests inherited production env vars AND the remote SurrealDB was reachable — which tracked exactly with factory agent sessions.

## Fix

Use `connect(&db_path, &SurrealConfig::default())` to force embedded mode, matching the pattern already established by `open_in_memory()`. +7/-2 lines.

## Test plan

- [x] `cargo test` — 676 passed, 0 failed, 3 ignored
- [x] `test_open_with_path` passed 10/10 consecutive runs (previously flaked ~1 in 5 on factory host)
- [x] Full suite clean

### DIFF
--- a/src/surreal_db/tests.rs
+++ b/src/surreal_db/tests.rs
@@ -14,7 +14,7 @@ fn test_schema_applies_without_error() {
 }
 
 #[test]
-fn test_open_with_path() {
+fn test_embedded_connect_creates_directory() {
     use super::connection::SurrealConfig;
     use tempfile::tempdir;
 
--- a/src/surreal_db/tests.rs
+++ b/src/surreal_db/tests.rs
@@ -15,13 +15,19 @@ fn test_schema_applies_without_error() {
 
 #[test]
 fn test_open_with_path() {
+    use super::connection::SurrealConfig;
     use tempfile::tempdir;
 
     let temp_dir = tempdir().unwrap();
     let db_path = temp_dir.path().join("test.surreal");
 
-    // Open database at specific path
-    let _db = SurrealDatabase::open(&db_path).unwrap();
+    // Use connect() with default config to force embedded mode, same as
+    // open_in_memory(). The public open() reads MX_SURREAL_MODE from the
+    // environment, which may route to a network backend that never creates
+    // a local directory -- making the exists()/is_dir() assertions flaky
+    // on hosts where the factory runs in network mode.
+    let config = SurrealConfig::default(); // always Embedded
+    let _db = SurrealDatabase::connect(&db_path, &config).unwrap();
 
     // Verify directory was created
     assert!(db_path.exists());


coryzibell:
## Code Review — Verdictia Stern

**No critical issues. No warnings. No required changes.**

### Verified
- `SurrealConfig::default()` forces embedded mode (`connection.rs:97-109`)
- `connect()` with a path creates the directory via `open_embedded_async` → `create_dir_all`
- Fix mirrors the `open_in_memory()` pattern exactly (`connection.rs:482-488`)
- No other tests have this vulnerability — all use `open_in_memory()` which already forces embedded
- Comment is accurate and proportionate

### S — Test name updated
`test_open_with_path` renamed to `test_embedded_connect_creates_directory` to reflect what it actually tests. Fixed in follow-up commit.

**Verdict:** Merge.

PR #301, opened 2026-05-08
kv: random entry sampling with constraints (#249)

### LINKED ISSUE #249: kv: random entry sampling with constraints
Add the ability to retrieve random entries from a history or list.

**Requested:**
- `mx kv random shipped --count 5` — get 5 random entries
- `mx kv random shipped --count 3 --since 30d` — 3 random from last 30 days
- `mx kv random ideas --count 1` — random idea for inspiration

Useful for reminiscing ("what did we do?"), random idea picks, or sampling large lists.

coryzibell:
Closes #249

## Summary

New `mx kv random` subcommand for random entry sampling from histories and lists. Supports `--count N` and time-range filtering. Also adds `--since <relative>` convenience flag to `TimeRangeArgs` (benefits `last`, `search`, `count`, and `random`).

## Files changed

- `src/cli.rs` — CLI argument definitions and subcommand wiring
- `src/kv.rs` — Core random sampling logic with constraint support
- `src/handlers/kv.rs` — Handler integration for the new subcommand

## Usage examples

```bash
mx kv random shipped --count 5
mx kv random ideas --count 1
mx kv random shipped --count 3 --since 30d
```

## Tests

- 8 new tests, 753 total passing
- Edge cases handled:
  - `count > available` returns all entries with a stderr note
  - Empty key returns empty result
  - Type mismatch errors cleanly for non-history/list types

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,14 +130,20 @@ mx kv last shipped --day 2026-04-25
 mx kv last shipped --month 2026-04
 mx kv last shipped --week 2026-W17
 mx kv last shipped --from 2026-04-01 --to 2026-04-15
+mx kv last shipped --since 1w
 mx kv search shipped "feature" --month 2026-04
 mx kv count shipped --day 2026-05-07
 
 # Time range composes with --count (filter first, then limit)
 mx kv last shipped --month 2026-04 --count 5
+
+# Random sampling from history/list keys
+mx kv random shipped --count 5
+mx kv random ideas --count 1
+mx kv random shipped --count 3 --since 30d
 ```
 
-Time-range flags (`--day`, `--month`, `--week`, `--from`/`--to`) are available on `last`, `search`, and `count`. All dates are UTC. For relative time queries (`1h`, `7d`), use `mx kv since`.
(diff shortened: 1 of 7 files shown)


coryzibell:
## Code Review

### Warnings

**W1 — Pre-existing: `last()` History branch doesn't guard empty timestamps; `random()` does.**
Not introduced by this PR. `random()` handles it correctly, making the asymmetry in `last()` more visible. Follow-up, not blocking.

**W2 — `count=0` silently accepted, returns empty vec.**
`--count 0` passes clap, returns nothing with no error. Almost certainly a user mistake. Fix: add `value_parser(clap::value_parser!(usize).range(1..))` to `--count`, or handle zero explicitly. No test covers this case.

### Suggestions

**S1 — Duplicated filter-then-warn block between History and List arms.**
Consistent with `last()` pattern — not blocking, but noted.

**S2 — No message when time range filters all entries.**
`count > available` prints a stderr note, but "all entries filtered out by time range" is silent. Consider a "note: no entries match the time range" message.

**S3 — `--since` accepts ISO-8601 strings unexpectedly.**
`parse_timeref` tries ISO-8601 first. Help text only mentions relative time. Either update the help text or restrict.

### Nits

**N1 — `count` parameter shadow in `sample` closure.** Consider `n` or `take`.

**N2 — Doc comment says "Get random entries" — missing the N.** Should be "Get N random entries from a history or list" to match `Last`.

### Verdict

No critical issues. Implementation is clean, pattern-consistent, tests are real. Fix W2 and the suggestions/nits.

PR #308, opened 2026-05-10
kv: push --create auto-adds key to schema (#253)

### LINKED ISSUE #253: kv push --create: auto-add key to schema if missing
## Summary

Currently if you `mx kv push` to a key that doesn't exist in the schema, it errors with `Unknown key`. You have to manually edit `queue.schema.toml` first, then push. Two steps for what should be one.

## Proposed

```bash
# Current (two steps):
# 1. edit schema.toml to add [keys.puns] with type/max_entries
# 2. mx kv push puns "the joke"

# Proposed (one step):
mx kv push puns "the joke" --create history --max-entries 500
```

If `--create` is passed and the key doesn't exist, auto-add it to the schema with the given type and options, then perform the push. If the key already exists, ignore `--create` and just push normally.

## Context

Discovered during Wake 355. Wanted to add a pun archive on the fly but had to stop, read the schema, add the key definition, save, then push. The schema is just TOML — appending a section is trivial.

coryzibell:
Closes #253

## Summary

`mx kv push <key> <value> --create <type>` auto-adds the key to the schema TOML if missing, then pushes. Supports `--max-entries`. Silently ignores `--create` if key already exists. Validates key names (no dots, special chars). Appends to TOML without reformatting.

## Files changed

- `src/cli.rs`
- `src/kv.rs`
- `src/handlers/kv.rs`

## Tests

- 11 new tests, 861 total passing

## Usage

```bash
mx kv push puns "the joke" --create history --max-entries 500
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,10 @@ mx kv inc builds
 mx kv push decisions "chose Typst for docs"    # prints: kv-A3fB (1)
 mx kv last decisions --count 5
 
+# Auto-create a key in the schema and push in one step
+mx kv push puns "the joke" --create history
+mx kv push ideas "wild thought" --create list --max-entries 500
+
 # Structured data on entries
 mx kv push projects "palmtop DSI fix" \
   --data '{"tags":["palmtop","i915"],"status":"active"}'
--- a/docs/src/architecture.typ
+++ b/docs/src/architecture.typ
@@ -609,6 +609,17 @@ files written before these fields existed are back-filled on first load
 (hashes are generated, data and memory default to `None`) and saved
 automatically.
 
+=== Schema mutation
+
+The `KvStore` struct holds a `schema_path` field alongside the existing
+`data_path`. The `add_key_to_schema()` method validates the key name
+(alphanumeric, underscores, hyphens; max 128 chars; no dots), appends a
+`[keys.<name>]` block to the TOML file without reformatting existing
+content, and re-parses the file to update the in-memory `Schema`. This is
+exposed through `push --create <type>` at the CLI layer, where the handler
+calls `add_key_to_schema` before the normal push path. If the key already
+exists, the method is a no-op.
+
 === Per-agent keying
 
 The active agent is determined by the `MX_CURRENT_AGENT` environment variable.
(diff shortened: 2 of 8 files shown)


coryzibell (commented):
## Review: kv push --create

Implementation is clean and well-structured. Reviewed all three changed files and the 11 new tests.

### What works well

- **CLI design**: `--create` as a `ValueEnum` enum means invalid types are rejected by clap before they reach `add_key_to_schema`. No need for a runtime type validation test -- the type system handles it.
- **Idempotency**: If the key already exists, `--create` is a silent no-op at both the handler level (schema check before calling `add_key_to_schema`) and the domain level (double-check inside the method). Safe to use unconditionally in scripts.
- **Schema preservation**: Append-only write to the TOML file avoids reformatting existing content. The test `add_key_to_schema_preserves_existing_content` verifies all original keys survive.
- **Key name validation**: Rejects dots (TOML quoting hazard), special characters, empty names, and names over 128 chars. Each rejection has a clear error message explaining *why*.
- **Error messages**: Every failure path includes the schema file path and the underlying OS error. The 2am reader will know exactly what went wrong.
- **Test coverage**: 11 tests covering the happy path (history, list, max_entries), idempotency (existing key noop), validation (dots, special chars, empty, too long, max length boundary), round-trip (reload from disk), content preservation, and the trailing-newline edge case.
- **`schema_path` on KvStore**: Clean addition -- the struct already had `data_path`, this is the natural complement.

### Observations (not blockers)

- The `add_key_to_schema` method accepts `value_type` as `&str` rather than `CreateType` or a domain enum. This works because the handler maps `CreateType` to the string before calling, and clap enforces the valid set. If a third push-compatible type is ever added, the handler mapping is the single place to update.
- The `let-else` chain in the handler (`if !store.schema.keys.contains_key(&key) && let Err(e) = ...`) uses a nightly-stabilized `let` chain pattern. Clean, but worth noting for anyone reading the code on an older Rust edition.

### Documentation

Added `--create` and `--max-entries` documentation across all five surfaces (kv.typ push section, kv.typ schema section, architecture.typ, getting-started.typ, README.md, index.typ). Docs build cleanly. Committed on this branch.

No blockers. Looks good.

PR #322, opened 2026-05-19
feat(kv): add update subcommand for in-place entry modification (#317)

gabaum10:
## From the Dumpster, With Love

The dumpster heard you. Issue #317 asked for a way to update an entry without losing its identity. We sent three opinionated trash pandas into a closed room to argue about what "update" should mean, then handed the verdict to a fourth raccoon to build it, then sent two more in to try to break it. Result: `mx kv update`. Preserves your `ts`, your hash, your position. Merges your `--data` patch field-by-field. Null in the patch deletes the field, because that's what you actually meant when you typed null. No `updated_at` — the issue allowed it, the raccoons said no, the spec sharpened that to a hard call.

## Who Built What

| Raccoon | Contribution |
|---------|-------------|
| **trashboat** | Won the parliament. Argued for merge semantics on day one — pointed out that replace forces re-supplying every required field on every partial update, which is worse than the remove+push workflow we were trying to replace. |
| **roach** | Came in fighting for replace, conceded by round 2 after trashboat's required-field argument. Then, post-build, found the test gap that mattered: the ambiguous-prefix unit test could silently skip its own assertion when random IDs didn't share a leading character. Inward lens, no infra-blaming. Mature. |
| **garbaggio** | The contrarian seat. Pushed back on null-as-skip vs null-as-delete and made the team think harder about it. Held the team's feet to the fire on the no-op rejection case. |
| **el-notario** | Solo builder. Two-pass borrow dance, closure-based merge, null-as-delete honored. 15 unit tests on the first pass, all GREEN. Then turned around and fixed both review warnings in 6 minutes. Decomposer mandate satisfied — the code reached the file. |
| **lil-grabby** | Pressure tested via the actual CLI binary. 22 adversarial tests in `tests/update_pressure.rs` — `--data '[1,2]'`, `--data 'null'`, empty-string values, all-null patches, sequential updates, you name it. Caught the empty-patch `{}` UX bug that nobody else thought to test. |

## DJ's Take

> Three raccoons walked into a debate room. One walked out with a different opinion than they walked in with. That's a good debate.

The shape was a debate-then-build run because the problem had real design space: merge vs replace on `--data`, preserve `ts` vs add `updated_at`, what to do when neither value nor data is provided. The parliament made the right call on all three: merge (so partial updates don't force re-supplying every required field), preserve `ts` (touching serialization for a field with no current consumer is scope creep), error on no-op (silent no-ops are debugging nightmares).

I sharpened one gap before the build: the proposal claimed "null deletes the field" but `validate_data`'s null-as-absent treatment was about validation behavior, not storage — leaving null in the JSON would have leaked through to downstream consumers. The spec made null-as-delete an active field removal during merge. The tests then proved this works (`update_data_null_deletes_field` removes the key from the object; `merge_resulting_in_empty_object_stored_as_none` normalizes empty merged objects back to `None`).

The review loop found two real warnings: `--data '{}'` slipped past the no-op guard (silent success — debugging hell), and the ambiguous-prefix test was probabilistic. Both fixed in commit `889d330`, with lil-grabby's pressure tests flipped to assert the new behavior — the tests themselves are the receipt that the fix landed.

## Test Results

977/977 GREEN.

- 955 unit tests in `src/kv.rs` (940 pre-existing + 15 new `update_*` tests)
- 22 integration tests in `tests/update_pressure.rs` (added during review)
- 3 ignored (unchanged, pre-existing)

## Stats

- 2 commits: `d0d1080` (feat), `889d330` (fix)
- 1491 insertions / 46 deletions across 4 files
- Run: 2026-05-19-175308
- First debate-mode run on this issue tracker

## Known Issues

- `_ =>` arm in `update_entry` returns `EntryNotFound` for the "key in schema but no entries pushed yet" case. Same pattern as `set_entry_memory`, so it's inherited, not introduced. If you want `KeyNotFound` semantics there, fix both sites in a follow-up.
- History and List arms in `update_entry` are structurally identical modulo `entries` vs `items`. ~90 lines of copy-paste. Could be a macro or a generic helper. Not in scope.
- Output format `Updated entry <index> (kv-<id>)` is human-friendly but different from push's machine-parseable `kv-<id> (<index>)`. Intentional per spec; flag if you want consistency.

---

*— DJ & the Trash Pandas*

*"We took your issue, argued about it for five minutes, fixed it in twelve, and reviewed it in eight. Your engineering org has nothing on a dumpster at 2am."*

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,8 @@
+# Increase the default stack size on Windows.
+#
+# Debug builds produce large stack frames due to unoptimised local variables.
+# The integration tests spawn `mx.exe` as a subprocess; with the default 1 MB
+# Windows stack the binary overflows on the deep `handle_kv` match dispatch.
+# 8 MB matches the stack size used by the Rust test harness on all platforms.
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "link-args=/STACK:8388608"]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1870,6 +1870,25 @@ pub enum KvCommands {
         all: bool,
     },
 
+    /// Update an existing entry's value and/or structured data in-place.
+    /// Preserves the entry's ID, position, and timestamp.
+    Update {
+        /// Key name
+        key: String,
+
+        /// New value text (omit to update only --data)
+        value: Option<String>,
+
+        /// Target entry by ID (numeric index or kv-HASH prefix). Required.
+        #[arg(long)]
+        id: String,
+
+        /// JSON object to merge into the entry's structured data.
+        /// Null field values delete that field from the merged result.
+        #[arg(long)]
+        data: Option<String>,
+    },
+
     /// Search entries in a list/history by substring and/or structured data filters
     Search {
         /// Key name
--- a/src/handlers/kv.rs
+++ b/src/handlers/kv.rs
@@ -725,6 +725,67 @@ pub(crate) fn handle_kv(cmd: KvCommands, verbose: bool) -> Result<i32> {
             }
         }
 
+        KvCommands::Update {
+            key,
+            value,
+            id,
+            data,
+        } => {
+            // 1. Reject the no-op case (both fields missing).
+            if value.is_none() && data.is_none() {
+                eprintln!("Error: provide a value argument and/or --data to update");
+                return Ok(kv::EXIT_INVALID_INPUT);
+            }
+
+            // 2. Parse the ID into IdRef (numeric or kv-HASH).
+            let id_ref = match parse_single_id(&id) {
+                Ok(r) => r,
+                Err(msg) => {
+                    eprintln!("Error: {}", msg);
+                    return Ok(kv::EXIT_INVALID_INPUT);
+                }
+            };
+
+            // 3. Parse --data as JSON object (mirror push's checks).
+            let parsed_data = match data {
+                Some(ref json_str) => {
+                    let val: serde_json::Value = match serde_json::from_str(json_str) {
+                        Ok(v) => v,
+                        Err(e) => {
+                            eprintln!("Error: invalid JSON for --data: {}", e);
+                            return Ok(kv::EXIT_INVALID_INPUT);
+                        }
+                    };
+                    if !val.is_object() {
+                        eprintln!(
+                            "Error: --data must be a JSON object, got {}",
+                            match val {
+                                serde_json::Value::Array(_) => "array",
+                                serde_json::Value::String(_) => "string",
+                                serde_json::Value::Number(_) => "number",
+                                serde_json::Value::Bool(_) => "boolean",
+                                serde_json::Value::Null => "null",
+                                serde_json::Value::Object(_) => unreachable!(),
+                            }
+                        );
+                        return Ok(kv::EXIT_INVALID_INPUT);
+                    }
+                    Some(val)
+                }
+                None => None,
+            };
+
+            // 4. Dispatch to engine.
+            match store.update_entry(&key, &id_ref, value.as_deref(), parsed_data) {
+                Ok(result) => {
+                    store.save()?;
+                    println!("Updated entry {} (kv-{})", result.index, result.id);
+                    Ok(kv::EXIT_OK)
+                }
+                Err(e) => handle_kv_err(e),
+            }
+        }
+
         KvCommands::Search {
             key,
             query,
--- a/src/handlers/kv.rs
+++ b/src/handlers/kv.rs
@@ -770,6 +770,10 @@ pub(crate) fn handle_kv(cmd: KvCommands, verbose: bool) -> Result<i32> {
                         );
                         return Ok(kv::EXIT_INVALID_INPUT);
                     }
+                    if val.as_object().is_some_and(|o| o.is_empty()) && value.is_none() {
+                        eprintln!("Error: --data is an empty object and no value was given — nothing to update");
+                        return Ok(kv::EXIT_INVALID_INPUT);
+                    }
                     Some(val)
                 }
                 None => None,
--- a/src/handlers/kv.rs
+++ b/src/handlers/kv.rs
@@ -771,7 +771,9 @@ pub(crate) fn handle_kv(cmd: KvCommands, verbose: bool) -> Result<i32> {
                         return Ok(kv::EXIT_INVALID_INPUT);
                     }
                     if val.as_object().is_some_and(|o| o.is_empty()) && value.is_none() {
-                        eprintln!("Error: --data is an empty object and no value was given — nothing to update");
+                        eprintln!(
+                            "Error: --data is an empty object and no value was given — nothing to update"
+                        );
                         return Ok(kv::EXIT_INVALID_INPUT);
                     }
                     Some(val)
--- a/src/kv.rs
+++ b/src/kv.rs
@@ -2273,52 +2273,61 @@ impl KvStore {
 
         // Compute merged data on a clone so we can validate before mutating.
         // This closure captures what we need from the entry without a mutable borrow.
-        let compute_merged = |entry_data: &Option<serde_json::Value>| -> Result<Option<serde_json::Value>, KvError> {
-            match (entry_data, &new_data) {
-                (_, None) => Ok(entry_data.clone()),
-                (None, Some(patch)) => {
-                    if !patch.is_object() {
-                        return Err(KvError::DataValidation {
-                            message: format!("key '{}': --data must be a JSON object", key),
-                        });
-                    }
-                    let mut obj = serde_json::Map::new();
-                    for (k, v) in patch.as_object().unwrap() {
-                        if !v.is_null() {
-                            obj.insert(k.clone(), v.clone());
-                        }
-                    }
-                    Ok(if obj.is_empty() { None } else { Some(serde_json::Value::Object(obj)) })
-                }
-                (Some(existing), Some(patch)) => {
-                    let mut merged_obj = match existing.as_object() {
-                        Some(o) => o.clone(),
-                        None => {
-                            return Err(KvError::Other(anyhow::anyhow!(
-                                "Data corruption: key '{}' entry has non-object data",
-                                key
-                            )));
-                        }
-                    };
-                    let patch_obj = match patch.as_object() {
-                        Some(p) => p,
-                        None => {
+        let compute_merged =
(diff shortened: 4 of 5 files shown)


coryzibell (commented):
## Review: feat(kv): add update subcommand for in-place entry modification (#317)

*reads the code three times*

---

### Critical Issues

**1. Two-pass lookup is a TOCTOU hazard with `.unwrap()` on the second pass** (`src/kv.rs`, both History and List arms of `update_entry`)

The code does a read-only first pass to find the entry and compute merged data, then a mutable second pass to apply the write. The second pass calls `.unwrap()` with the comment "safe: existence confirmed above."

This is safe *today* because both passes happen within a single `&mut self` call with no interleaving. But it is fragile by design. The `unwrap()` makes it a panic-on-invariant-violation rather than a recoverable error. If this code is ever refactored to use async, interior mutability, or if the entry lookup logic between passes ever drifts, the process crashes instead of returning an error.

The correct fix is to use indexed access. On the first pass, record the *position* (vec index) of the matched entry. On the second pass, access by position. No second lookup, no `unwrap()`. This would also eliminate the duplicated `IdRef` match blocks entirely.

For example:
```rust
let pos = entries.iter().position(|e| e.index == *n)
    .ok_or_else(|| KvError::EntryNotFound { ... })?;
let merged = compute_merged(&entries[pos].data)?;
def.validate_data(key, &merged)?;
entries[pos].value = ...;
entries[pos].data = merged;
```

This is not blocking, but it is the difference between code that is safe by proof and code that is safe by coincidence.

**2. Massive code duplication between History and List arms** (`src/kv.rs`, the entire `update_entry` method)

The History and List arms are ~60 lines each and are structurally identical. The only difference is the field name (`entries` vs `items`). This is the same pattern already present in `set_entry_memory`, so it is a pre-existing sin, but this PR doubles down on it rather than addressing it.

If the lookup-by-ID logic, the merge logic, or the validation call ever needs a fix, it must be applied in two places. That is how bugs happen.

Not blocking for this PR since it follows the existing pattern, but it should be tracked as tech debt.

---

### Warnings

**3. `_ =>` catch-all returns misleading `EntryNotFound` error** (`src/kv.rs`, bottom of `update_entry`)

When a key exists in the schema as `history` or `list`, but no data has been pushed yet, `self.data.entries.get_mut(key)` returns `None`. This falls through to the `_ =>` arm, which returns `KvError::EntryNotFound`. The error message will say something like `Entry not found: ID Index(0) in key 'log'`.

This is technically correct (the entry does not exist), but the `_ =>` arm also catches `DataValue::Counter`, `DataValue::String`, and `DataValue::State` -- types that should have been rejected by the type check at the top of the method. If there is data corruption (schema says `history` but stored data is `counter`), this arm silently returns `EntryNotFound` instead of surfacing the corruption.

A cleaner approach: match explicitly on `Some(DataValue::Counter { .. }) | Some(DataValue::String { .. }) | Some(DataValue::State { .. })` and return a different error, and handle `None` separately with a more descriptive message like "no entries exist for key 'log'."

Low severity since the type check at the top should prevent the corruption case, but defense in depth matters.

**4. The handler rejects `--data '{}'` with no value, but `update_entry` itself does not** (`src/handlers/kv.rs` vs `src/kv.rs`)

The empty-data-no-value rejection lives only in the handler layer (line 84 of the diff in `handlers/kv.rs`). The engine method `update_entry` would happily accept `Some(json!({}))` with `None` value and produce a no-op write. Any future caller of `update_entry` (API, different handler, tests) would not get this guard.

The validation should live in the engine, not the handler. Push it into `update_entry` at the top, next to the type check.

**5. `data` parameter in the handler is `ref`-bound unnecessarily** (`src/handlers/kv.rs`, line 62 of the diff)

```rust
Some(ref json_str) => {
```

The `data` variable is `Option<String>` owned by the match arm. The `ref` is unnecessary here because `data` is already bound and `json_str` could just be obtained via `data.as_ref()` or by destructuring without `ref`. This is not wrong, but it is a code smell that suggests the author was fighting the borrow checker rather than understanding the ownership. Minor.

---

### Suggestions

**6. The `compute_merged` closure captures `&new_data` by reference, which complicates lifetimes unnecessarily**

The closure `compute_merged` is defined inside `update_entry` and captures `&new_data`. It is called exactly once. This could be a plain function or a local helper that takes `new_data` as an explicit parameter, making the data flow clearer. As a closure, a reader has to scan upward to figure out what it captures.

**7. Test file naming: `tests/update_pressure.rs`**

The file exists in the diff but not on disk on `main` -- confirmed it is new in this PR. The name `update_pressure` is good: it conveys adversarial/stress testing. The 20 tests are thorough and cover the important edge cases: empty patch, null-as-delete, required field protection, type mismatch rollback, ambiguous prefix, empty string value, output format verification.

One gap: there is no test for updating an entry in a List with `--data` (the List+data path). All data-merge tests use History. Given the code duplication concern in point 2, this is exactly the kind of path that could silently break. Add at least one integration test that exercises List + `--data` merge.

**8. Unit tests in `kv.rs` are solid**

The 15 unit tests cover: value-only update (history and list), data merge (history and list), null-as-delete, combined value+data, validation rejection (unknown field, type mismatch, required field via merge), not-found by index and prefix, ambiguous prefix, wrong key type, and prefix lookup. Good coverage. The `test_schema_with_data` helper is clean.

**9. The `UpdateResult` struct could carry more information**

`PushResult` has `index` and `id`. `UpdateResult` has `index` and `id`. They are structurally identical. Consider whether they should be the same type (e.g., `EntryRef` or `EntryResult`), or whether `UpdateResult` should also carry what changed (value, data, or both) for downstream consumers. Minor, but worth considering for API ergonomics.

---

### Assessment

The feature is correct. The update logic properly preserves entry ID, position, and timestamp. The shallow merge with null-as-delete is implemented correctly. Schema validation runs on the merged result before committing the write -- this is the right ordering. Error handling surfaces appropriate exit codes through the existing `handle_kv_err` pathway.

The test coverage is strong: 15 unit tests + 20 integration tests covering the important edge cases.

The main structural concern is the duplicated History/List arms and the two-pass unwrap pattern. Neither is a correctness bug today, but both are maintenance liabilities.

Fix the `--data '{}'` guard location (move it to the engine). The rest is acceptable.

*the smallest nod*

Fix items 3 and 4. The rest can ship.

coryzibell (changes requested):
## 🥜 Sorting Room Verdict — Changes Requested

The update logic is sound and the tests are solid. Four items need addressing before this ships:

### Critical

**1. Double-pass `.unwrap()` in `update_entry` (src/kv.rs)**
The method finds the target entry via immutable iteration, then finds it *again* via mutable iteration and calls `.unwrap()`. This is safe by coincidence — the entry was just proven to exist — but not by construction. If any intermediate operation ever mutates the vec between passes, this panics in production. Use positional indexing instead: find the index, then mutate `entries[idx]` directly.

**2. History/List code duplication (src/kv.rs)**
~60 lines of near-identical logic between History and List variant handling. The PR doubles down on existing duplication rather than extracting the shared pattern. Extract a common update path that both variants call.

### Warnings (still requesting fixes)

**3. Catch-all `_ =>` returns misleading `EntryNotFound` (src/kv.rs)**
The match arm swallows states that aren't actually "entry not found" — could be data corruption, an empty key with no entries, or a variant the code doesn't expect yet. Return a more specific error or distinguish the cases.

**4. Empty-data guard lives in the handler, not the engine (src/handlers/kv.rs)**
The `--data {}` rejection check is in the CLI handler layer. Any future caller of `update_entry` (tests, other commands, library use) bypasses this validation entirely. Move the guard into the engine method itself.

### CI

**5. Fix Windows CI test failures.** The build needs to pass on all platforms before merge.

---

*The squirrels sorted every nut. These need re-roasting before they go in the bar.*

coryzibell (approved):
## Re-review: 6 Original Findings

Checked each against commits 55dfcfa3 and e610fe20.

### 1. CRITICAL: Double-pass `.unwrap()` in `update_entry` — RESOLVED

The `find_entry_idx` helper now returns a positional `usize` index into the slice. `apply_entry_update` uses `entries[idx]` for the immutable data read, then `&mut entries[idx]` for the mutation. No `.unwrap()`, no double lookup. Clean.

### 2. CRITICAL: History/List code duplication — RESOLVED

The `UpdateableEntry` trait abstracts the shared field access. `find_entry_idx` and `apply_entry_update` are generic over `T: UpdateableEntry`, called once for History entries and once for List items. Zero duplication in the update logic. The trait impls are mechanical but necessary — acceptable boilerplate.

### 3. WARNING: Catch-all `_ =>` returns misleading `EntryNotFound` — RESOLVED

The `TypeMismatch` guard now runs *before* the data match, so the `_ =>` arm only fires when `entries.get_mut(key)` returns `None` (key exists in schema but no data written yet). In that case `EntryNotFound` is semantically correct — there genuinely are no entries. The data-corruption scenario (schema says History, stored data is Counter) remains theoretically misleading, but it cannot occur through normal operation and does not warrant its own error variant. Acceptable.

### 4. WARNING: Empty-data guard in handler only — RESOLVED

The engine method `update_entry` at line 2437 now checks for the empty-object-with-no-value case and returns `KvError::DataValidation`. Library callers get the same guard as the CLI. The handler *also* checks (defense in depth) — that's fine.

### 5. FIX: Unnecessary `ref`-binding on `data` — NOT FIXED

`src/handlers/kv.rs` line 751: `Some(ref json_str)` in the Update handler. The `ref` is unnecessary — `data` is not used after the match consumes the inner string. However, this is consistent with the existing Push handler at line 466, which uses the same pattern. The `ref` is harmless — it borrows instead of moving, which compiles identically when the owned value is not used afterward. Noting for completeness but not blocking.

### 6. FIX: Windows CI test failures — RESOLVED

`.cargo/config.toml` increases the stack size to 8 MB for `x86_64-pc-windows-msvc` via linker flags. The comment explains *why*: debug builds produce large stack frames, and the deep `handle_kv` match dispatch overflows the default 1 MB Windows stack. This is a legitimate fix, not a skip/ignore. Scoped to Windows MSVC only. Reasonable.

---

## New Code Quality

The fix commits are clean. No regressions introduced. The `UpdateableEntry` trait is well-designed — minimal surface area, clear semantics. The `apply_entry_update` function handles all merge paths (None/None, None/Some, Some/Some) with proper null-as-delete semantics and validation-before-write.

Test coverage is thorough: 15 unit tests in `kv.rs`, 20 integration tests in `update_pressure.rs` covering edge cases (ambiguous prefixes, required field deletion, empty patches, type mismatches, sequential updates, empty strings, output format verification).

5 of 6 findings resolved. The remaining `ref`-binding (#5) is cosmetic and consistent with existing code.

*the nod*

PR #323, opened 2026-05-20
fix: skip invalid JSONL lines in codex save (#178)

gabaum10:
## 🦝 From the Dumpster

Raccoon fix for codex save failing on invalid JSONL lines (issue #178). DJ escape-hatched the debate, one builder (lil-grabby), three review iterations caught a half-applied fix. Clean after iter 3.

---
*— DJ & lil-grabby* 🦝
*The parliament stayed home. The review loop didn't.*

### DIFF
--- a/src/codex/archive/write.rs
+++ b/src/codex/archive/write.rs
@@ -352,7 +352,10 @@ pub(crate) fn archive_session(
                 let agent_content = match fs::read_to_string(&source_path) {
                     Ok(c) => c,
                     Err(e) => {
-                        eprintln!("warning: failed to read agent file {:?}: {}", source_path, e);
+                        eprintln!(
+                            "warning: failed to read agent file {:?}: {}",
+                            source_path, e
+                        );
                         continue;
                     }
                 };
--- a/src/codex/archive/write.rs
+++ b/src/codex/archive/write.rs
@@ -349,16 +349,28 @@ pub(crate) fn archive_session(
         if !agents.is_empty() {
             for agent in &agents {
                 let source_path = PathBuf::from(&agent.id);
-                if let Ok(agent_content) = fs::read_to_string(&source_path)
-                    && let Ok((_modified_agent_content, agent_images, agent_stats)) =
-                        extract_images_from_jsonl(&agent_content, &images_dir)
-                {
-                    total_stats.parsed += agent_stats.parsed;
-                    total_stats.skipped += agent_stats.skipped;
-                    for img in agent_images {
-                        if !all_images.iter().any(|existing| existing.hash == img.hash) {
-                            all_images.push(img);
+                let agent_content = match fs::read_to_string(&source_path) {
+                    Ok(c) => c,
+                    Err(e) => {
+                        eprintln!("warning: failed to read agent file {:?}: {}", source_path, e);
+                        continue;
+                    }
+                };
+                match extract_images_from_jsonl(&agent_content, &images_dir) {
+                    Ok((_modified_agent_content, agent_images, agent_stats)) => {
+                        total_stats.parsed += agent_stats.parsed;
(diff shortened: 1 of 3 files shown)


coryzibell (approved):
## Verdict

*silence*

*reads*

*reads again*

*reads a third time*

This is a clean fix. The approach is correct: skip and warn, don't abort. The implementation matches the intent. Three files touched, all three make sense.

### What's right

**The skip-and-continue pattern.** `serde_json::from_str` fails, the line gets counted as skipped, an eprintln warning fires with line number and a 50-char escaped preview, and processing continues to the next line. The bad line is dropped from the output JSONL — correct. Propagating corruption into the archive would be worse than dropping it.

**The all-corrupt bailout.** If every non-empty line fails to parse, the function returns an error instead of silently producing an empty archive. This is the right boundary: one bad line is corruption you can survive; a file of nothing but bad lines is a file that isn't what you think it is.

**Empty input handling.** Empty string and all-blank-lines both return successfully with zero stats. No false bailout on empty content. Correct.

**The stats aggregation in write.rs.** Session stats and agent stats are accumulated separately then summed. The "skipped > 0" conditional in the summary output means users only see the parsed/skipped breakdown when something was actually dropped. Clean.

**The agent error handling improvement** (second commit). The first commit used chained `if let Ok(...)` which silently swallowed both file-read errors and extraction errors. The second commit breaks these into explicit match arms with eprintln warnings. Better. Errors should be visible.

**Visibility tightening.** `pub(crate)` to `pub(super)` on `JsonlParseStats` and `total()`. Correct — nothing outside the codex module needs these.

**The `format_skip_warning` extraction.** Deduplicates the warning format between `count_images_in_jsonl` and `extract_images_from_jsonl`. The test (`warning_includes_one_indexed_line_number`) validates the exact format string. Good.

**Test coverage.** 12 tests covering: valid lines, mixed valid/invalid, all-invalid (error), empty content, blank lines, null bytes (the actual issue #178 trigger), multibyte UTF-8 in preview, line number indexing, output exclusion of bad lines. The edge cases that matter are covered.

**migrate.rs changes.** Mechanical tuple-destructuring updates (`_stats`). The migration code doesn't use the stats — that's fine. Migration is a batch operation; the per-line warnings still fire via eprintln.

### Notes (not blocking)

1. **Line number semantics across files.** When warnings fire for agent files, the line numbers are local to that agent file, but there's no indication of *which* file the warning came from. The `format_skip_warning` only says "line N". In write.rs, the agent-file error handler prints the path, but only if `extract_images_from_jsonl` returns `Err` (the all-corrupt case). For the partial-skip case where the function returns `Ok` with `skipped > 0`, the individual per-line warnings from inside `extract_images_from_jsonl` don't include the filename. In practice this is fine — stderr context from the surrounding code usually makes it clear — but if you ever want to make these warnings grep-friendly, adding the filename to `format_skip_warning` would help.

2. **The `preview_truncates_to_50_chars` test** asserts `p.len() == 50` for ASCII input. This works because `escape_default` is identity for printable ASCII. For non-ASCII input (the multibyte test), the assertion only checks `!p.is_empty()`. The preview function takes 50 *chars* then escapes, so the byte length of the output can exceed 50. That's the right behavior for a warning message — just noting that the "50 char" claim is about input chars, not output bytes.

3. **No test for partial JSON lines** (e.g., `{"type":"user"` — a truncated JSON object). These will fail `serde_json::from_str` and get skipped, which is correct. Not a gap, just noting the coverage boundary.

*the smallest nod*

The fix is surgical. The tests cover the edges. The second commit addressed real review findings. Ship it.

coryzibell (changes requested):
Don't merge this until the changes to 322 have been made and that is approved and merged. Then rebase this onto main

PR #327, opened 2026-05-21
feat(kv): batch multi-field set for state entries (#324)

coryzibell:
## Summary

Two new input modes for `mx kv set` on state-type entries:

- **Inline `key=value` pairs** — trailing varargs after the entry name (e.g. `mx kv set mystate foo=bar baz=42`)
- **`--json` flag** — accepts a JSON object for state entries or a JSON array for tensor entries

Both modes are atomic (all-or-nothing), support partial updates (unmentioned fields are preserved), and are schema-validated against the entry's declared schema.

## What changed

- **Engine** (`kv.rs`): 2 new methods — `set_state_fields` for batch field writes and `set_json` for JSON-based input
- **CLI** (`cli.rs`): trailing varargs support + `--json` flag on `kv set`
- **Handler** (`kv_handler.rs`): `SetInput` enum with parsing logic to disambiguate legacy single-field, inline multi-field, and JSON input modes

## Backward compatibility

Legacy `mx kv set <key> <field> <value>` syntax continues to work unchanged.

## Test coverage

25 new tests (12 engine, 13 handler) — 968 total passing.

Closes #324

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,12 @@ mx kv inc builds
 mx kv push decisions "chose Typst for docs"    # prints: kv-A3fB (1)
 mx kv last decisions --count 5
 
+# Batch set multiple state fields at once
+mx kv set context goal="done" phase="writing"
+mx kv set context --json '{"goal":"done","phase":"writing"}'
+mx kv set mytensor --json '[0.4, 0.6, 0.5]'
+echo '{"goal":"done"}' | mx kv set context --json -
+
 # Auto-create a key in the schema and push in one step
 mx kv push puns "the joke" --create history
 mx kv push ideas "wild thought" --create list --max-entries 500
--- a/docs/src/architecture.typ
+++ b/docs/src/architecture.typ
@@ -589,7 +589,7 @@ Supported types:
   [`string`], [Simple string value. Supports `set`, `get`],
   [`history`], [Timestamped append-only log with optional `max_entries` cap. Supports `push`, `last`, `since`, `search`, `count`, `random`. Each entry gets a numeric index and a stable base58 entry ID (`kv-` prefix). Entries can carry optional structured JSON data (`--data` on push, `--where` on queries). The `last`, `search`, `count`, and `random` commands accept time-range flags (`--day`, `--month`, `--week`, `--since`, `--from`/`--to`) for date filtering.],
   [`list`], [Ordered list with timestamps. Supports `push`, `pop`, `remove`, `search`, `count`, `random`. Each entry gets a numeric index and a stable base58 entry ID. Entries can carry optional structured JSON data. The `last`, `search`, `count`, and `random` commands accept the same time-range flags as history.],
-  [`state`], [Named fields (like a struct). Supports `set <key> <field> <value>`, `get`],
+  [`state`], [Named fields (like a struct). Supports single-field set (`set <key> <field> <value>`), batch set (`set <key> field=value ...` or `set <key> --json '{...}'`), tensor positional set (`set <key> --json '[...]'`), and `get`. Batch operations validate all fields against the schema before writing.],
 )
 
 === Data (JSON)
--- a/docs/src/getting-started.typ
+++ b/docs/src/getting-started.typ
@@ -145,6 +145,12 @@ mx kv get decisions --id kv-A3fB              # look up by entry ID
 mx kv push puns "the joke" --create history
 mx kv push ideas "wild thought" --create list --max-entries 500
 
+# Batch set multiple state fields at once
+mx kv set context goal="done" phase="writing"
+mx kv set context --json '{"goal":"done","phase":"writing"}'
+mx kv set mytensor --json '[0.4, 0.6, 0.5]'
+echo '{"goal":"done"}' | mx kv set context --json -
+
 # Link entries to the memory graph
 mx kv push decisions "adopted memory links" --memory kn-abc123
 mx kv set decisions --id 17 --memory kn-abc123
--- a/docs/src/kv.typ
+++ b/docs/src/kv.typ
@@ -156,17 +156,32 @@ KV commands use structured exit codes for scripting:
 )
 
 #command(
-  "mx kv set <key> <value> [field_value]",
+  "mx kv set <key> [args...] [--json <value>]",
   [Set a value for a string, counter, or state key, or link a specific
-  entry to a memory node.
+  entry to a memory node. Supports four input modes for state keys:
+  single-field, inline batch, JSON object, and JSON array (tensor).
 
   For *string* keys: `mx kv set <key> <value>` sets the value directly.
 
   For *counter* keys: `mx kv set <key> <value>` parses the value as an integer
   and clamps to min/max.
 
-  For *state* keys: `mx kv set <key> <field> <value>` sets a single field.
-  The field name must be declared in the schema.
+  For *state* keys, four input modes are available:
+
+  / Single field (legacy): `mx kv set <key> <field> <value>` -- sets one field. Backward compatible with pre-batch syntax.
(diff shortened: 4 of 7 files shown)


coryzibell (commented):
## Code Review — `feat/batch-state-set`

*Three reads. Here is the verdict.*

---

### [W1] `BatchFields` and `JsonObject` skip `--memory` fallback logic

**handlers/kv.rs, ~lines 579-596 (dispatch block)**

The `Scalar` and `SingleField` arms both have the careful `--memory` fallback: if the set fails with `TypeMismatch` but `--memory` was provided, it prints a warning and continues to the memory-pointer section. The `BatchFields`, `JsonObject`, and `JsonArray` arms do not — they `return handle_kv_err(e)` unconditionally.

This is probably intentional (batch is state-only, so TypeMismatch means "you used batch on a non-state key" which is a real error), but it is an observable behavior difference. If someone passes `mx kv set mystring goal=done --memory kn-abc`, they get a hard exit instead of the memory-only fallback they would have gotten with the old two-arg form. Document the intent or unify the behavior.

**Severity: W** — users combining `--memory` with batch on a wrong-typed key get a worse experience than before.

---

### [W2] `parse_positional_args` empty-key-name in `key=value` is silently accepted

**handlers/kv.rs, lines 233-235 (1-arg `=` case), 253-254 (2-arg), 267-268 (3+ arg)**

`split_once('=')` on `"=hello"` yields `("", "hello")`. An empty field name passes through to `set_state_batch`, where it gets validated against the schema — but only if the schema has a `fields` list. A state key without an explicit `fields` schema would silently accept `""` as a field name. This is a data quality hazard.

Add `if k.is_empty() { return Err(...) }` in `parse_positional_args` immediately after `split_once`.

**Severity: W** — bad data can enter the store if schema is lax.

---

### [W3] JSON nested objects silently stringify via `other.to_string()`

**handlers/kv.rs, lines 305-308 (JSON object value coercion)**

For JSON object values, `String` and `Number` are handled explicitly, but everything else (`Object`, `Array`, `Bool`, `Null`) falls through to `other.to_string()`. This means `{"goal": {"sub": "task"}}` silently stores the field value as the literal string `{"sub":"task"}`. Booleans become `"true"`/`"false"`, null becomes `"null"`.

This is not necessarily wrong — all KV state fields are strings — but it is surprising. A user passing `--json '{"active": true}'` likely expects a boolean, not the string `"true"`. At minimum, document this coercion. Better: reject non-string/non-number values in JSON objects and tell the user to stringify deliberately.

Same pattern exists in the array arm (lines 322-327) where `other.to_string()` silently stringifies nested objects/arrays/bools/nulls in the array.

**Severity: W** — silent data coercion is a readability problem for the user's mental model.

---

### [S1] `SetInput::BatchFields` and `SetInput::JsonObject` are structurally identical

**handlers/kv.rs, lines 210-212**

Both are `Vec<(String, String)>`. Both dispatch to `store.set_state_batch(&key, &pairs)`. The handler arms at lines 579-596 are copy-pasted. The only reason they exist as separate variants is provenance tracking (positional vs JSON). But that provenance is never used — not in error messages, not in logging, not in tests.

Collapse into a single `BatchFields { pairs: Vec<(String, String)>, source: InputSource }` or just `BatchFields(Vec<(String, String)>)` and let the caller not care. This removes 6 lines of duplicated dispatch.

**Severity: S** — the code is correct, but the enum lies about the distinction mattering.

---

### [S2] `set_tensor_batch` double-validates type

**kv.rs, lines 1219-1228 and then delegates to `set_state_batch` at line 1260**

`set_tensor_batch` checks `def.value_type != ValueType::State` and returns `TypeMismatch`, then calls `set_state_batch` which checks the same thing again. The second check always passes (or the first would have caught it), but it is wasted work and the `.clone()` of the key def happens twice.

Consider making `set_tensor_batch` build the pairs and then call `set_state_batch` directly without the type guard (since `set_state_batch` has the same guard). Or extract the common validation into a shared helper.

**Severity: S** — redundant work, not a bug.

---

### [S3] No test for `--json` + positional args mutual exclusion in the handler

**handlers/kv.rs, lines 507-510**

The runtime guard `if json.is_some() && !args.is_empty()` is good, but there is no unit test covering it. The parser tests cover `parse_positional_args` and `parse_json_input` individually, but the handler-level guard that prevents combining them is untested. Consider an integration test or at minimum a handler-level unit test.

**Severity: S** — the guard is trivial, but untested guards are untrustworthy guards.

---

### [S4] No test for JSON with boolean/null/nested-object values

**handlers/kv.rs tests, lines 1377-1440**

There is `json_object_with_numeric_value` but no test for `{"active": true}`, `{"cleared": null}`, or `{"nested": {"a": 1}}`. Given the `other.to_string()` coercion discussed in W3, these are exactly the cases that need test coverage to make the behavior explicit and intentional.

**Severity: S** — the coercion behavior should be pinned by tests regardless of whether you keep or reject it.

---

### [N1] `trailing_var_arg = true` appears twice

**cli.rs, lines 1727 and 1734**

Both `#[command(trailing_var_arg = true)]` on the `Set` variant and `#[arg(trailing_var_arg = true)]` on the `args` field. Clap only needs one (on the command). The field-level annotation is redundant. Not harmful, just noise.

**Severity: N**

---

### [N2] Comment says "atomic operation" in `set_state_batch` doc

**kv.rs, line 1125**

"Single atomic operation" is accurate in the sense that all validation happens before any in-memory writes, and `save()` does atomic file rename. But the word "atomic" has strong connotations (transactions, rollback). The real guarantee is "validate-all-then-write-all" with atomic file persistence. The doc is fine but could be more precise.

**Severity: N**

---

### What is correct

- **Backward compatibility is preserved.** The 2-arg-no-equals path (`mx kv set key field value`) correctly routes to `SingleField`, which calls the existing `store.set()`. Good.
- **Atomicity is real.** `set_state_batch` validates all field names before touching the in-memory state. If any field is invalid, nothing is written. `save()` uses tmp+fsync+rename. The batch is genuinely all-or-nothing.
- **Duplicate field detection in batch.** The `HashSet` approach is correct and reports all duplicates.
- **Error reporting batches all schema violations.** Multiple unknown fields are all reported in one error. This is the right choice.
- **`split_once('=')` for values containing `=`.** Correctly splits on first `=` only, preserving `key=value` in the value portion. Tested.
- **`set_tensor_batch` delegates to `set_state_batch`.** Clean composition. Exact-count enforcement is clear.
- **Exit codes are consistent.** `DataValidation` maps to `EXIT_INVALID_INPUT` (4). Type mismatches map to `EXIT_TYPE_MISMATCH` (2). Correct.
- **Test coverage is solid for the happy paths and the main error paths.** 22 new tests across engine and handler layers.

The parsing heuristic (equals-sign detection) is about as good as it can be given the backward-compatibility constraint. The two-bare-args ambiguity is inherent and the chosen resolution (legacy wins) is the right call.

---

### Verdict

Fix **W1** (decide on `--memory` fallback behavior for batch arms) and **W2** (reject empty field names in the parser). The rest are improvements, not blockers.

*The nod is close. Not yet.*

coryzibell (commented):
Re-review complete. All eight findings verified. Verdict: **APPROVE** (posting as comment due to self-review restriction).

- **W1**: BatchFields and JsonArray arms both have the --memory TypeMismatch fallback. Matches Scalar/SingleField pattern.
- **W2**: Empty field name rejection in all three branches (1-arg, 2-arg, 3+). Tests cover each.
- **S1**: JsonObject variant removed. parse_json_input returns BatchFields for objects. Enum is clean.
- **S2**: set_tensor_batch delegates to set_state_batch for type validation. No duplication.
- **S3**: Mutual exclusion test present.
- **S4**: Coercion tests pin boolean→"true", null→"null", integer→"42".
- **N1**: Single trailing_var_arg on Set variant. No duplicates.
- **N2**: Doc comment on set_state_batch covers all six behaviors.

975 tests pass, 0 failures. One cosmetic note: stale doc comment at line 293 of handlers/kv.rs still says "route to JsonObject" — not blocking.

The verdict: this is correct.

PR #330, opened 2026-05-23
feat: embed full documentation in binary via mx docs (#329)

coryzibell:
## Summary

- Adds `mx docs` subcommand that prints the full mx documentation as Markdown to stdout
- Documentation is zstd-compressed at build time via `build.rs` and embedded in the binary with `include_bytes!`
- `mx help` now shows a hint: "Run `mx docs` for full documentation as Markdown"

## Pipeline

```
docs/build.sh         build.rs              mx docs
  (pandoc)        (zstd compress)          (zstd decompress)
     |                  |                       |
  *.typ --> mx-docs.md --> mx-docs.md.zst --> stdout
```

## Changes

- **`build.rs`** (new): Reads `docs/out/llm/mx-docs.md`, compresses with zstd via base-d, writes blob to `$OUT_DIR`
- **`src/embedded.rs`** (new): `include_bytes!` the compressed blob, decompress on demand
- **`src/cli.rs`**: Added `Docs` variant to `Commands` enum and `after_help` hint
- **`src/main.rs`**: Added `mod embedded` and `Commands::Docs` match arm
- **`Cargo.toml`**: Added `base-d = "3"` as build dependency
- **`.gitignore`**: Added negation for `docs/out/llm/mx-docs.md` so it's tracked in git (the build.rs depends on it existing at build time)
- **`docs/out/llm/mx-docs.md`**: Now tracked in git (7442 lines, ~245KB source, compresses to ~50KB in the binary)

## Test results

- `cargo run -- docs | head -20` prints the start of the markdown
- `cargo run -- docs | wc -l` returns 7442 (matches source file)
- `cargo run -- help` shows the `mx docs` hint
- `cargo run -- help kv` still works normally
- `cargo test` passes all 1006+ tests
- `cargo clippy` passes with zero warnings

Closes #329

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,8 +64,8 @@ $RECYCLE.BIN/
 *.iml
 *.ipr
 *.iws
-# out/ -- disabled; conflicts with docs/out/ negation.
-# IntelliJ output dirs are already covered by .idea/
+# out/ -- disabled; a bare out/ rule here would shadow the docs/out/
+# negation above (gitignore is order-dependent, last matching rule wins).
 .idea_modules/
 
 # Vim
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,11 @@
 **/*.rs.bk
 *.pdb
 
-# Docs build output
-docs/out/
+# Docs build output (keep LLM markdown for build.rs embedding)
+docs/out/*
+!docs/out/llm/
+docs/out/llm/*
+!docs/out/llm/mx-docs.md
 
 # macOS
 .DS_Store
@@ -61,7 +64,8 @@ $RECYCLE.BIN/
 *.iml
 *.ipr
 *.iws
-out/
+# out/ -- disabled; conflicts with docs/out/ negation.
+# IntelliJ output dirs are already covered by .idea/
 .idea_modules/
 
 # Vim
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,9 @@ colored = "2"
 # Embeddings
 fastembed = { version = "5.6", default-features = false, features = ["hf-hub-rustls-tls", "image-models", "ort-download-binaries-rustls-tls"] }
 
+[build-dependencies]
+base-d = "3"
+
 [dev-dependencies.cargo-husky]
 version = "1"
 default-features = false
--- a/build.rs
+++ b/build.rs
@@ -15,8 +15,8 @@ fn main() {
     let raw = fs::read(src)
         .expect("docs/out/llm/mx-docs.md must exist at build time — run docs/build.sh first");
 
-    let compressed =
-        compress(&raw, CompressionAlgorithm::Zstd, ZSTD_COMPRESSION_LEVEL).expect("zstd compression failed");
+    let compressed = compress(&raw, CompressionAlgorithm::Zstd, ZSTD_COMPRESSION_LEVEL)
+        .expect("zstd compression failed");
 
     let out_dir = env::var("OUT_DIR").unwrap();
     let dest = Path::new(&out_dir).join("mx-docs.md.zst");
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,19 @@
+use std::env;
+use std::fs;
+use std::path::Path;
+
+use base_d::{CompressionAlgorithm, compress};
+
+fn main() {
+    let src = Path::new("docs/out/llm/mx-docs.md");
+    println!("cargo:rerun-if-changed={}", src.display());
+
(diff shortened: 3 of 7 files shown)


coryzibell (commented):
## Code Review: `feat: embed full documentation in binary via mx docs (#329)`

---

### build.rs

**[S] Compression level is hardcoded — name it.** Line 15: `compress(&raw, CompressionAlgorithm::Zstd, 3)`. The `3` is the zstd default level, which is fine, but a named constant communicates intent better than a magic number. Minor.

**[N] The `cargo:rerun-if-changed` only watches the single file.** This is correct for the current setup — `mx-docs.md` is the sole input. If the build pipeline ever starts reading multiple files, this will need updating, but that's a future concern, not a present one.

**API usage is correct.** `compress()` signature matches: `(&[u8], CompressionAlgorithm, u32) -> Result<Vec<u8>, Box<dyn Error>>`. The `expect()` on the source read is appropriate — build scripts *should* fail hard with a clear message when a required input is missing. The message tells the developer exactly what to do. Good.

---

### embedded.rs

**Correct.** `include_bytes!` with `concat!(env!("OUT_DIR"), ...)` is the standard pattern. `decompress()` signature matches: `(&[u8], CompressionAlgorithm) -> Result<Vec<u8>, Box<dyn Error>>`. The two `expect()` calls are justified — both conditions (corrupt blob, invalid UTF-8) mean the binary itself is broken. The doc comment explains this reasoning. The module-level doc comment is present.

The nod.

---

### cli.rs

**Line 15: `after_help` text is clean.** `"Run \`mx docs\` for full documentation as Markdown"` — good discoverability, appropriate placement.

**Line 146: `Docs` variant.** Unit variant, no arguments, descriptive doc comment. Placed last in the enum, after all other commands. Fits the existing pattern.

---

### main.rs

**Line 9: `mod embedded;`** — placed alphabetically in the module list. Correct.

**Lines 90-93: Match arm.** `print!` (not `println!`) is the right call — the markdown file already has its own trailing newline. Clean pattern: no unnecessary binding, no error path beyond what `docs_markdown()` handles internally.

---

### Cargo.toml

**[build-dependencies] `base-d = "3"`** — matches the runtime dependency at line 45 (`base-d = "3"`). Same major version, same spec. Cargo will resolve to the same version. Correct.

---

### .gitignore

**The negation pattern is correct:**
```
docs/out/*
!docs/out/llm/
docs/out/llm/*
!docs/out/llm/mx-docs.md
```
This is the standard git negation dance: ignore everything in `docs/out/`, except the `llm/` directory, then ignore everything in `llm/`, except `mx-docs.md`. The intermediate `!docs/out/llm/` is required because git won't look inside an ignored directory. Textbook.

**[W] Removal of bare `out/` rule.** The old IntelliJ `out/` rule was removed and replaced with a comment explaining why. The comment claims "IntelliJ output dirs are already covered by .idea/". This is *mostly* true — IntelliJ's default `out/` directory for compiled output is separate from `.idea/` and is NOT covered by ignoring `.idea/`. However: no `out/` directories exist in this repo outside of `docs/out/` (verified), and IntelliJ/CLion for Rust projects use `target/` (already ignored). The practical risk is near zero. The comment could be more precise — the real reason is that a bare `out/` rule at line 67 conflicts with the `docs/out/` negation pattern above it, since gitignore rules are order-dependent and a later `out/` would re-ignore `docs/out/llm/mx-docs.md`. The current comment doesn't explain *why* it conflicts, just that it does.

---

### docs/out/llm/mx-docs.md

7,442 lines, 250KB. This is the compiled documentation checked into the repo. It's a build artifact, but it's intentionally tracked so that `cargo build` works without requiring `docs/build.sh` to have been run first. The `.gitignore` negation pattern ensures only this specific file is tracked. Reasonable tradeoff — the file is human-readable markdown, diffs are meaningful, and it prevents a build-time dependency on Typst/Pandoc tooling.

---

### Edge cases considered

**Fresh clone without `docs/build.sh`:** The file is checked into git. `cargo build` will find it. If someone deletes it locally without running `docs/build.sh`, `build.rs` will panic with a clear message. This is the correct behavior.

**File changes:** `cargo:rerun-if-changed` is set. Cargo will re-run `build.rs` when the markdown changes. If it doesn't change, the compressed blob is reused. Correct.

---

### Summary

No critical issues. One warning (the `.gitignore` comment could be more precise about *why* `out/` conflicts). Two minor items. The implementation is clean, the API usage is verified correct against the base-d source, the patterns are idiomatic Rust, and the tradeoffs are well-reasoned.

**Verdict: APPROVE.**

PR #331, opened 2026-05-23
kv schema: support description field for keys

coryzibell:
## Summary

Closes #316.

- Adds an optional `description: Option<String>` field to `KeyDef` in `src/kv.rs`, following the existing `Option<String>` pattern used by `default`
- Updates `keys()` return type to `Vec<(&str, ValueType, Option<&str>)>` so callers receive the description
- Updates the `kv keys` handler to display descriptions as a third column when present; keys without descriptions print exactly as before
- Fixes all destructuring sites (`resolve_dump_memories`, manual `KeyDef` constructors in tests)
- Adds `key_description_round_trips` test verifying parse and retrieval

## Test plan

- [x] `cargo test` -- all 1005+ unit tests and 22 integration tests pass
- [x] `cargo clippy -- -D warnings` -- clean
- [x] New test `key_description_round_trips` covers: schema with description parses, `keys()` returns it, schema without description returns `None`

### DIFF
--- a/docs/src/architecture.typ
+++ b/docs/src/architecture.typ
@@ -574,6 +574,7 @@ default = ""
 
 [keys.focus_areas]
 type = "list"
+description = "Areas of active focus"
 
 [keys.session_state]
 type = "state"
--- a/docs/src/kv.typ
+++ b/docs/src/kv.typ
@@ -71,6 +71,7 @@ type = "list"
 [keys.todos]
 type = "list"
 max_entries = 20
+description = "Pending work items"
 
 [keys.context]
 type = "state"
@@ -84,6 +85,7 @@ Schema fields:
 / `min`: Optional. Minimum value for counters (clamped, never errors).
 / `max`: Optional. Maximum value for counters (clamped, never errors).
 / `max_entries`: Optional. Maximum entries for history and list types. Oldest entries are dropped when exceeded. Omit to allow unbounded growth.
+/ `description`: Optional. Human-readable description of the key's purpose. Displayed as a third column by `mx kv keys`.
 / `fields`: Optional. List of valid field names for state types. Writes to unlisted fields are rejected.
 
 === Auto-creating keys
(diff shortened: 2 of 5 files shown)


coryzibell (commented):
## Review: feat(kv): support description field for keys (#316)

*three reads complete*

---

### [W] Ragged column alignment in `Keys` handler — `src/handlers/kv.rs` lines 1189-1190

```rust
Some(d) => println!("{:30} {:10} {}", name, vtype, d),
None => println!("{:30} {}", name, vtype),
```

When some keys have descriptions and others do not, the `vtype` column is 10-wide on description rows and unpadded on no-description rows. If a schema mixes keys with and without descriptions, the output columns will not align vertically.

Fix: use `{:10}` for `vtype` on both branches, or better, always use the three-column format and leave the description empty:

```rust
match desc {
    Some(d) => println!("{:30} {:10} {}", name, vtype, d),
    None => println!("{:30} {:10}", name, vtype),
}
```

This keeps the type column consistent regardless of whether a description is present.

---

### [S] Consider `#[serde(skip_serializing_if = "Option::is_none")]` — `src/kv.rs` line 204

`KeyDef` only derives `Deserialize`, not `Serialize`, so this is a non-issue today. But if `Serialize` is ever added, the `description: None` fields will serialize as `description = ""` in TOML or `"description": null` in JSON. Mentioning it only because every other `Option` field on this struct has the same latent concern — not specific to this PR.

No action required.

---

### What is correct

- **Schema parsing.** `description: Option<String>` with `#[serde(default)]` correctly makes the field optional. Absent keys parse as `None`. Present keys parse as `Some(String)`. Standard pattern, matches every other optional field on `KeyDef`.

- **Return type.** `Vec<(&str, ValueType, Option<&str>)>` is the right shape. `v.description.as_deref()` correctly converts `Option<String>` to `Option<&str>` to borrow from the schema. All four call sites (`resolve_dump_memories`, `Keys` handler, two tests) are updated.

- **`resolve_dump_memories` destructuring.** `(key, _vtype, _desc)` — correct. The function only needs the key name for memory lookup.

- **Tests.** `keys_lists_all` verifies all test_schema keys have `None` descriptions. `key_description_round_trips` verifies TOML deserialization with both present and absent descriptions. Both tests cover the right boundaries.

- **Literal `KeyDef` constructions.** All three (`sample_key_def`, `key_def_with_data`, `validate_data_none_passes_freeform`) include `description: None`. No missing fields.

- **No `--json` mode on `Keys`.** Confirmed `Keys` is a unit variant with no flags. No JSON output path to update.

---

### Verdict

One warning: the column alignment inconsistency. Everything else is clean.

Fix the alignment. The rest gets the nod.

PR #332, opened 2026-05-23
feat(memory): auto-anchor re-evaluation and --no-auto-anchor flag (#199)

### LINKED ISSUE #199: auto_anchor: add --no-auto-anchor opt-out flag and visibility output for cleanup batches
# auto_anchor: re-evaluate existing anchors on each run

## The friction

`auto_anchor` (`mx/src/main.rs:2227-2339`) runs synchronously on every memory write — Add/Update/Edit/Append/Prepend. It computes cosine similarity against all entries with embeddings, takes top 5 (`max_anchors = 5`, similarity threshold 0.75-0.95), and adds them to the entry's anchor list. **It only ADDS — never removes.**

For normal usage this works great — the graph self-strengthens as entries get touched. But two friction points emerge over time:

1. **Monotonic accumulation.** As the corpus evolves, old anchors become stale (the entries that were once semantically similar drift apart) but stay in the list forever. The graph slowly fills with edges that no longer reflect current relationships.

2. **Cleanup batches inflate the graph.** A cleanup operation that touches N entries (e.g., re-typing or re-categorizing) inflates the graph by up to N×5 anchor edges as a side effect, even when the cleanup intent is metadata-only. The embedding text doesn't include category or resonance, so a category-only update produces an identical embedding vector — but `auto_anchor` still re-runs the similarity search and writes whatever new neighbors it finds.

We hit this in our environment running memory graph hygiene work. Loam (our memory infrastructure agent) attributed the +5 inflation to "phantom concurrent processes" for two sessions before we traced it to `auto_anchor`. The mechanism is correct, but the cumulative effect is real.

Filing background context: the issue surfaced for us during a calibration follow-up session in the Soren environment. We're not arguing the design is wrong — we love `auto_anchor` for normal usage. Just proposing a small refinement that keeps the graph clean by construction.

## The proposal

Have `auto_anchor` re-evaluate existing anchors on every run, dropping stale ones automatically:

1. For each anchor already in the entry's list, recompute current cosine similarity against the entry's embedding
2. If it's dropped below the 0.75 threshold, drop the anchor
3. Then add new top-5 from non-anchored neighbors (the existing logic)
4. Merge → dedup → upsert (the existing tail end)

**Net effect:** anchors live as long as they're still semantically similar. Stale ones get pruned automatically. The cleanup-batch problem disappears as a side effect — even if a cleanup batch touches many entries, the total anchor count stays bounded because old anchors get dropped as new ones arrive. The graph stays clean by construction, not by manual hygiene.

**Cost:** Extra cosine computation per existing anchor on every write. For an entry with N existing anchors, that's N additional comparisons. Probably negligible at typical anchor counts (single digits to low tens), but worth measuring at scale. If it matters, the optimization is straightforward (cache the entry's embedding vector for the comparison loop instead of refetching each anchor's embedding from the DB — most of the cost is the I/O, not the math).

**Optional safety net:** Combine with a hard cap on total anchors per entry (e.g., max 20). If the cap would be exceeded after re-eval + new additions, drop the lowest-similarity ones. Bounded growth as a backstop. Not required, just a defense-in-depth option if the re-eval logic ever has a corner case.

## Offer

Happy to PR this if you'd like — the change is small and additive to the existing `auto_anchor` function structure. Or take it yourself, whatever works. No rush; we have a hygiene workaround on our side (snapshot anchor counts before bulk operations and document the inflation), this is the underlying fix.

coryzibell:
## Summary

- **Stale anchor pruning**: `auto_anchor` (in `helpers.rs`) and the `AutoAnchor` batch command (in `handlers/memory.rs`) now re-evaluate existing anchors. Anchors whose cosine similarity has drifted below the threshold (0.75) or above the near-duplicate ceiling (0.95) are pruned from the entry's anchor list.
- **`--no-auto-anchor` flag**: Added to six mutation subcommands (Add, Update, Edit, Append, Prepend, Restore) in `cli.rs`. When set, skips the automatic anchor generation step and prints `(auto-anchor skipped)` to stderr. Does NOT skip `auto_embed` -- only anchoring is affected.
- **Batch visibility**: The `AutoAnchor` batch command now reports stale anchor pruning counts alongside new anchor additions, with `x` prefix for removals and `+` prefix for additions.

## Files changed

- `src/cli.rs` -- `no_auto_anchor: bool` field added to six subcommand structs
- `src/helpers.rs` -- `auto_anchor()` restructured: stale anchor detection, combined early-return guard, filtered anchor rebuild with `explicitly_removed` integration
- `src/handlers/memory.rs` -- Six handler destructuring patterns updated; `auto_anchor` calls wrapped in `if !no_auto_anchor`; batch `AutoAnchor` command gets same stale-pruning logic

## Test results

- `cargo clippy`: zero warnings
- `cargo test`: 1004 unit + 22 integration tests pass, 0 failures

Closes #199

### DIFF
--- a/docs/src/memory.typ
+++ b/docs/src/memory.typ
@@ -69,6 +69,7 @@ similarity, and surfaced through keyword or semantic search.
     ([`--type`],       [`string`], [Fact type for ephemeral knowledge: `decision`, `insight`, `person`, `quote`, `thread_opened`, `commitment`, `thread_closed`. Auto-routes category and sets `resonance_type=ephemeral`.]),
     ([`--session`],    [`string`], [Session to link fact to via EXTRACTED_FROM relationship. Requires `--type`.]),
     ([`--thread-id`],  [`string`], [Thread ID for `thread_closed` operations. Requires `--type`.]),
+    ([`--no-auto-anchor`], [`flag`], [Skip automatic anchor generation.]),
     ([`--json`],       [`flag`],   [Output as JSON.]),
   ),
   examples: (
@@ -223,6 +224,7 @@ to have embeddings generated via `mx memory embed`.]
     ([`--owner`],         [`string`], [Update owner (only valid when visibility is private).]),
     ([`--session-id`],    [`string`], [Update session ID (for retrofitting entries with wrong or missing session linkage).]),
     ([`--force`],         [`flag`],   [Force dangerous visibility changes (e.g., making blooms public).]),
+    ([`--no-auto-anchor`], [`flag`],  [Skip automatic anchor generation.]),
     ([`--json`],          [`flag`],   [Output as JSON.]),
   ),
   examples: (
@@ -243,6 +245,7 @@ to have embeddings generated via `mx memory embed`.]
     ([`--replace`],     [`string`], [Replacement text. Also accepts `--new`.]),
     ([`--replace-all`], [`flag`],   [Replace all occurrences (default: error if multiple matches).]),
     ([`--nth`],         [`int`],    [Replace only the Nth occurrence (1-indexed).]),
+    ([`--no-auto-anchor`], [`flag`], [Skip automatic anchor generation.]),
     ([`--json`],        [`flag`],   [Output as JSON.]),
   ),
   examples: (
@@ -258,6 +261,7 @@ to have embeddings generated via `mx memory embed`.]
   flags: (
     ([`--content`], [`string`], [Content to append (omit to read from stdin).]),
     ([`-f, --file`], [`path`],  [Read content from file. Also accepts `--content-file`.]),
+    ([`--no-auto-anchor`], [`flag`], [Skip automatic anchor generation.]),
     ([`--json`],    [`flag`],   [Output as JSON.]),
   ),
   examples: (
@@ -273,6 +277,7 @@ to have embeddings generated via `mx memory embed`.]
   flags: (
     ([`--content`], [`string`], [Content to prepend (omit to read from stdin).]),
     ([`-f, --file`], [`path`],  [Read content from file. Also accepts `--content-file`.]),
+    ([`--no-auto-anchor`], [`flag`], [Skip automatic anchor generation.]),
     ([`--json`],    [`flag`],   [Output as JSON.]),
   ),
   examples: (
@@ -286,6 +291,7 @@ to have embeddings generated via `mx memory embed`.]
   before restoring.],
   flags: (
     ([`--list`], [`flag`], [List available backups instead of restoring.]),
+    ([`--no-auto-anchor`], [`flag`], [Skip automatic anchor generation.]),
     ([`--json`], [`flag`], [Output as JSON.]),
   ),
   examples: (
@@ -408,7 +414,12 @@ Anchors are connections between entries discovered via embedding similarity.
 #command(
   "mx memory auto-anchor",
   [Automatically add anchors between entries based on embedding similarity.
-  Processes a single entry or all entries that have embeddings.],
+  Processes a single entry or all entries that have embeddings.
+
+  Also re-evaluates existing anchors: any anchor whose cosine similarity has
+  fallen below the threshold (default 0.75) or risen above the near-duplicate
+  ceiling (0.95) is pruned. This keeps the anchor graph self-cleaning --
+  anchors that made sense once but no longer do are removed automatically.],
   flags: (
     ([`--threshold`],   [`float`], [Minimum cosine similarity (0.0--1.0). Default: `0.75`.]),
     ([`--max-anchors`], [`int`],   [Maximum anchors to add per entry. Default: `5`.]),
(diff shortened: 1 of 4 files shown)


coryzibell (commented):
## Code Review -- Verdictia Stern

*silence*

*reads*

*reads again*

*reads a third time*

---

### Critical

None.

### Warnings

**1. Missing `total_pruned` counter in `AutoAnchor` batch summary** (`src/handlers/memory.rs`, around line 1883)

The per-entry output reports "Added N anchors, pruned M stale" but the final summary only reports `total_added`. There is no `total_pruned` accumulator. A batch run that prunes 50 stale anchors across 200 entries ends with a summary that says nothing about pruning. The per-entry lines scroll off. The summary is the only thing that sticks.

Fix: add `let mut total_pruned = 0;` alongside `total_added`, increment it next to `total_added += top_matches.len();`, and update the summary to include it.

**2. Magic number `0.95` appears four times without a name** (`src/helpers.rs:280`, `src/helpers.rs:314`, `src/handlers/memory.rs:1775`, `src/handlers/memory.rs:1800`)

The near-duplicate ceiling is hardcoded as `0.95` in four locations across two files. In the `AutoAnchor` batch command, `threshold` is a named CLI parameter but the ceiling is a bare literal. If someone adjusts the ceiling in one place and misses the others, the staleness check and the new-anchor check diverge silently.

Fix: extract to a named constant. `const NEAR_DUPLICATE_CEILING: f32 = 0.95;` in helpers.rs, and reference or pass it into the batch handler.

### Suggestions

**3. New `explicitly_removed` filter in `helpers.rs` (lines 340-342) is redundant for the current call flow**

The update handler upserts the entry (with anchors already reflecting `--anchors` removal) *before* calling `auto_anchor`. So when `auto_anchor` fetches the entry from the DB, `entry.anchors` no longer contains the removed anchors, making the `retain` call at line 340-342 a no-op. This is harmless -- belt-and-suspenders -- but it would benefit from a comment explaining *why* it exists (defense against future callers that might not pre-upsert). Without that comment, the next reviewer will spend time tracing the same flow I just did.

**4. Output symbol change in batch handler** (`src/handlers/memory.rs:1835-1838`)

The output symbol changed from `→` to `+` for new anchor matches. This is a reasonable change (distinguishes additions from pruning's `x`), but it is a silent breaking change for any tooling that parses stdout. If nothing parses this output, it is fine. If something does, it needs coordination.

### Verdict

The pruning logic is correct. The threshold and ceiling boundary conditions are consistent between the stale check and new-anchor check. The privacy filter is correctly NOT re-applied to existing anchors during re-evaluation. The `explicitly_removed` interaction is sound. The `no-op` guard correctly requires BOTH stale_anchors and similarities to be empty. The `--no-auto-anchor` flag reaches all six handler call sites with correct destructuring.

The flag propagation is clean. The conditional pattern (`if !no_auto_anchor { ... } else { eprintln!(...) }`) is repeated identically six times -- repetitive but consistent, which I will take over clever abstraction any day.

Fix the `total_pruned` counter (warning 1). The magic number (warning 2) is not urgent but should be addressed before the ceiling value ever needs to change.

*the smallest nod*

The rest is acceptable.

coryzibell (commented):
## `--fill` flag review — Verdictia Stern

*silence*

*reads*

*reads again*

*reads a third time*

---

### Findings by severity

**No critical issues.**

**No warnings.**

**Suggestions (2):**

**1. `--fill` with a specific `id` — silent no-op deserves a message.**

`mx memory auto-anchor some-entry-id --fill` where `some-entry-id` already has anchors will print the fill mode banner, print "Processing 1 entries...", then skip the entry silently (or with a verbose message only), then print the summary showing 0 added, 0 pruned. The user gets no clear signal that their specific entry was skipped *because of `--fill`*.

This is not wrong. But it is confusing. A user who passes both a specific ID and `--fill` is probably confused about what `--fill` means, or forgot the entry already has anchors. A one-line message outside verbose mode — something like `"Skipped {id} (already has anchors, --fill)"` — would save them a head-scratch. Low priority.

**2. Dry-run summary does not report `--fill` skipped count.**

The `--fill` skip report (`"Skipped N entries that already had anchors"`) is inside the `else` branch of the dry-run check (line 1908). So `--dry-run --fill` prints `[DRY RUN] Complete. No changes written.` but never reports how many entries were skipped. The skip information is only visible via `--verbose`. This is a minor gap — someone previewing with `--dry-run --fill` probably wants to see the skip count too. Move the skip report outside the `if dry_run` / `else` block so it prints in both paths.

---

### What's correct

- **CLI definition.** Flag is clean. Help text is clear. No `conflicts_with` needed — `--fill` composes naturally with every other flag.
- **Skip placement.** The `continue` is at the top of the loop, before any similarity computation. Correct — no wasted work.
- **`entry.anchors.is_empty()` check.** Correct predicate for the stated semantics. Entries with stale-but-present anchors are still considered "anchored" and skipped. This is the right behavior — `--fill` means "fill gaps in the graph," not "ensure quality." The pruning path is for `auto-anchor` without `--fill`.
- **Counter tracking.** `total_skipped` increments correctly, `total_added` and `total_pruned` are untouched for skipped entries. Math is clean.
- **`--no-auto-anchor` on write commands.** Consistent across all six mutation paths. Pattern is identical in each: guard, else-branch prints skip message. Mechanical but correct.
- **Constants extraction.** `NEAR_DUPLICATE_CEILING` and `DEFAULT_ANCHOR_THRESHOLD` pulled into `helpers.rs` with doc comments. Shared between `auto_anchor` helper and `AutoAnchor` handler. Good.
- **Stale anchor pruning.** The broader pruning logic (not `--fill` specific) is sound — re-evaluates existing anchors against the threshold and ceiling, removes them in both the batch handler and the per-entry helper.
- **Documentation.** The `memory.typ` additions are accurate and match the implementation.

---

Two suggestions, both minor output polish. The logic is correct.

*the smallest nod*

Issue #2, opened 2025-11-28
Add dejavu detection to upload commit


coryzibell:
Detect when both title and body randomly receive the same dictionary from base-d.

This is observation, not forcing - watching the system for coincidence. When dejavu occurs, add decode hint to footer.

PR #335, opened 2026-05-24
feat: replace fastembed/ort with tract-onnx for pure-Rust embeddings (#334)

coryzibell:
## Summary

- **Removed** the `fastembed` dependency (which pulled in `ort`, the C++ ONNX Runtime bindings) and replaced it with `tract-onnx` + `tokenizers` + `hf-hub` -- all pure Rust
- **Rewrote** `src/embeddings.rs`: `FastEmbedProvider` is now `TractProvider`, which loads BGE-Base-EN-v1.5 via `hf-hub`, tokenizes with the `tokenizers` crate, and runs inference through `tract-onnx` with CLS pooling and L2 normalization
- **Renamed** `fastembed_cache_dir()` to `model_cache_dir()` in `src/paths.rs`, with backward-compatible env var support (`MX_ISOLATE_MODELS` new, `MX_ISOLATE_FASTEMBED` legacy -- both honored)
- **Updated** all references in `src/helpers.rs` and `src/handlers/memory.rs`

## Why

The `fastembed` crate depends on `ort`, which downloads and links against the C++ ONNX Runtime. This creates build complexity (C++ toolchain requirements, platform-specific binary downloads) and increases binary size. `tract-onnx` is a pure-Rust ONNX runtime that produces identical embeddings (verified by the PoC in `experiments/tract-poc/`).

## Dependency changes

| Removed | Added |
|---------|-------|
| `fastembed 5.6` (+ `ort`, C++ ONNX Runtime) | `tract-onnx 0.22` |
| | `tokenizers 0.21` (with `fancy-regex`) |
| | `hf-hub 0.4` (with `ureq` for sync downloads) |

## What stays the same

- The `EmbeddingProvider` trait interface -- unchanged
- The model: `BAAI/bge-base-en-v1.5` (768-dim) -- same model, same embeddings
- Existing embeddings in the database -- remain valid
- The `build.rs` -- untouched (it's for docs compression, not fastembed)

## Test results

- `cargo test`: 1006 + 22 tests pass, 0 failures
- `cargo clippy -- -D warnings`: clean
- `cargo tree | grep -i "ort\|fastembed"`: no matches (C++ dep chain fully removed)

Closes #334


coryzibell (commented):
## Code Review — `feat: replace fastembed/ort with tract-onnx for pure-Rust embeddings`

---

### Critical

None.

---

### Warnings

**1. `embed()` clones and recompiles the model on every call (embeddings.rs:85-98)**

Every invocation of `embed()` clones the `InferenceModel`, sets input facts, runs `into_optimized()`, and `into_runnable()`. For single calls this is fine. For `embed_batch` with N texts, you pay the clone+optimize+compile cost N times. The old `fastembed` library batched properly.

This matters because `mx memory embed --all` iterates every entry in the store. With hundreds of entries, you are recompiling the ONNX graph hundreds of times.

Acceptable for v1 -- the comment on line 142 acknowledges this is the simple approach. But a follow-up should cache compiled models by sequence length (`HashMap<usize, TypedRunnableModel>`) to avoid redundant compilation. Flag it as a TODO so it doesn't get lost.

**2. `model_cache_new_env_takes_precedence_over_legacy` test name and comment contradict behavior (paths.rs:592-599)**

The test is named `model_cache_new_env_takes_precedence_over_legacy`. The comment says "New env var (empty = not isolated) should take precedence over legacy." But the assertion expects `home.join("memory").join("embed")` -- the *isolated* path -- because the empty new env var is filtered out and the legacy env var takes effect.

The test *passes* and the *code* is correct. But the name and comment are misleading. If the new env var truly took precedence, an empty value would mean "not isolated" and the legacy `Some("1")` would be ignored, producing the XDG path. Instead, empty is treated as absent, and legacy wins. The test proves backward compat, not precedence.

Rename to something like `model_cache_empty_new_env_falls_through_to_legacy`.

**3. Cross.toml still references ort-sys (Cross.toml:2)**

The comment on line 2 says `ort-sys v2.0.0-rc.10 build-dependency uses ureq with native-tls (openssl)`. Since ort-sys is gone, this comment is stale. Furthermore, with pure-Rust tract-onnx, the OpenSSL pre-build steps in Cross.toml may no longer be needed (unless another dependency still requires them -- check). At minimum, update the comment. At best, test cross-compilation without the pre-build and remove it if it passes.

**4. `docs/out/llm/mx-docs.md` has 10+ stale FastEmbed references**

Lines 6179-6181, 6237, 6501, 6628, 6700, 6828, 6862, 7069, 7076-7077 all reference `FastEmbed`, `FastEmbedProvider`, `fastembed_cache_dir`, and the fastembed crate. This is generated documentation but it describes the old architecture. If it is auto-generated, regenerate it. If hand-maintained, update it.

---

### Suggestions

**5. CLS pooling extraction could be simpler (embeddings.rs:125-128)**

```rust
let mut cls_pooled = vec![0.0f32; hidden_size];
for hidden_idx in 0..hidden_size {
    cls_pooled[hidden_idx] = output_tensor[[0, 0, hidden_idx]];
}
```

This allocates a zero-filled vec then overwrites every element. Simpler:

```rust
let cls_pooled: Vec<f32> = (0..hidden_size)
    .map(|i| output_tensor[[0, 0, i]])
    .collect();
```

Or better, use ndarray slicing directly if tract_ndarray exposes it:

```rust
let cls_pooled = output_tensor.slice(s![0, 0, ..]).to_vec();
```

Not a correctness issue. Readability.

**6. `embed_batch` with empty input returns `Ok(vec![])` (embeddings.rs:141-144)**

This is correct behavior -- `iter().map().collect()` on empty input produces an empty vec. Worth noting it is intentional, not accidental, with a brief comment. The old fastembed may have had different behavior with empty input.

**7. Model ID inconsistency: `Xenova/bge-base-en-v1.5` vs `BAAI/bge-base-en-v1.5` (embeddings.rs:38, 60)**

The HF Hub fetch uses `Xenova/bge-base-en-v1.5` (the ONNX-exported variant) but `model_id()` returns `BAAI/bge-base-en-v1.5` (the original model). This is semantically correct -- the BAAI ID identifies the *model*, Xenova identifies the *ONNX export* -- but anyone debugging embedding mismatches will see `BAAI/...` in the database and not find the actual HF repo. Consider a comment explaining the distinction, or storing both.

**8. Magic number `3` in input fact loop (embeddings.rs:89)**

```rust
for i in 0..3 {
```

This assumes the model has exactly 3 inputs (input_ids, attention_mask, token_type_ids). Correct for BGE-base-en-v1.5, but fragile if the model is ever swapped. Consider:

```rust
let input_names = ["input_ids", "attention_mask", "token_type_ids"];
for (i, name) in input_names.iter().enumerate() { ... }
```

Or at minimum, a comment explaining the magic 3.

---

### Verdict

The core implementation is correct. tract API usage, tokenization, CLS pooling, L2 normalization -- all sound. The trait is satisfied. The old dependency is fully excised from Rust source. Error handling propagates properly throughout.

The per-call model recompilation (#1) is the only thing with real runtime cost, and it is acceptable for v1 with a documented follow-up. The stale docs (#4) and Cross.toml (#3) need cleanup before or shortly after merge.

Fix #2 (test name), #3 (Cross.toml comment), and #4 (docs). The rest can land as-is.

coryzibell (commented):
## Re-review -- verifying fixes for previous findings

*reads*
*reads again*
*reads a third time*

### Previous findings -- status

| ID | Finding | Verdict |
|----|---------|---------|
| W2 | Test rename to match actual behavior | **Fixed.** `model_cache_empty_new_env_falls_through_to_legacy` -- name accurately describes the test. |
| W3 | Cross.toml stale ort-sys comment | **Fixed.** Now references hf-hub and tract-onnx. Accurate. |
| W4 | Doc sources updated, docs rebuilt | **Fixed.** Both `architecture.typ` and `paths.typ` updated. `mx-docs.md` rebuilt and consistent with sources. |
| S5 | CLS pooling idiomatic ndarray slicing | **Fixed.** Line 127 of `embeddings.rs`: `output_tensor.slice(tract_ndarray::s![0, 0, ..]).to_vec()` -- proper `s!` macro usage. |
| S7 | Xenova vs BAAI model ID comment | **Fixed.** Lines 38-40 of `embeddings.rs` explain the distinction clearly. |
| S8 | Comment explaining 3 model inputs | **Fixed.** Line 92 of `embeddings.rs` names all three inputs. |

All six previous findings addressed correctly.

### New issues found on third read

**[W1] `architecture.typ` version mismatch with `Cargo.toml`**

`Cargo.toml` line 66: `tract-onnx = "0.22"`
`docs/src/architecture.typ` line 51: `[0.21]`
`docs/out/llm/mx-docs.md` line 6630: `0.21`

The docs say 0.21, the actual dependency is 0.22. The docs were rebuilt from stale source. Fix the version in `architecture.typ` and rebuild.

**[S1] `README.md` still lists `MX_ISOLATE_FASTEMBED` as a current env var**

`README.md` line 256: `MX_KV_SCHEMA`, `MX_KV_DATA`, `MX_ISOLATE_FASTEMBED`.

This should say `MX_ISOLATE_MODELS` (with an optional note that the legacy name is still honored). The README is user-facing -- it should advertise the current name.

Note: the `MX_ISOLATE_FASTEMBED` references in `paths.rs` (backward-compat code, legacy doc comments, test comments) and in `paths.typ`/`mx-docs.md` (documenting the legacy fallback) are all correct and intentional. Only the README is stale.

---

Fix the version in W1 and the README in S1. The rest is acceptable.

Issue #250, opened 2026-04-26
kv: ID range and multi-ID queries


coryzibell:
Add the ability to query entries by ID range or multiple specific IDs.

**Requested:**
- Single ID: `mx kv get shipped --id 35`
- ID range: `mx kv get shipped --id 35-64`
- Multiple IDs: `mx kv get shipped --id 1,5,12,35`

Currently you can only get entries by `last --count N` or `search`. No way to grab specific entries by their numeric ID or a range of them.

PR #375, opened 2026-06-09
kv: introduce 'kv schema' subcommand suite (list/add/drop/update); fold in 'kv keys' (#363)

coryzibell:
Closes #363

## Summary

Consolidates KV schema lifecycle management under a single `mx kv schema [list | add | drop | update]` namespace. Closes the orphaned-key gap (no removal path existed) and makes key definition a first-class verb for all five types. The legacy top-level `kv keys` and `kv rename` verbs are retained as deprecated aliases — no breaking changes this cycle.

## Commands

- `mx kv schema list` — absorbs `kv keys`; stdout byte-identical.
- `mx kv schema add <key> --type <T>` — first-class definition for all five `ValueType`s, with `--max-entries / --default / --min / --max / --description / --fields / --data`. Routes through the same engine as the retained `push --create` shortcut.
- `mx kv schema drop <key> [--force]` — removes schema definition **and** stored data. `--force` gates non-empty keys; silent for empty/never-written. Unregistered → `KeyNotFound`.
- `mx kv schema update <key>` — safe metadata edits only (`--description / --max-entries / --min / --max / --add-field`); `--name` delegates to `rename_key`. `--type` is forbidden and points at `kv migrate` (no override).

## Binding decrees honored (kn-fdabe4c6 / kn-14ed7e8a)

1. `update` retype — **forbidden**, errors to `kv migrate`, no `--force`.
2. Rename — top-level `kv rename` kept working AND mirrored as `schema update --name`; both route to `rename_key`. Top-level marked deprecated (clap doc-comment + stderr notice).
3. `drop` non-empty — `--force` required; silent for empty/never-written.
4. `drop` unregistered — error via `KvError::KeyNotFound`.
5. No `--json` flag anywhere (deferred to the mx-wide JSON overhaul).
6. `kv keys` deprecated alias of `schema list` — stderr notice, stdout byte-identical.
7. `drop`/`update` follow the `rename_key` transaction pattern (mutate in-memory, persist data-first then schema, roll back on data-write failure).

## Spec ambiguity resolved (flagged for review)

The issue prose says unregistered `drop` returns **"exit code 3 (KeyNotFound)"**, but the canonical KV exit-code table (decree kn-daf2020b, PR #250) maps `key-not-found = 1` and `schema-missing = 3`. Every other verb returns **exit 1** for an unknown key. Honoring the decree's "consistent with every other verb" intent, I routed `drop` of an unregistered key through `KvError::KeyNotFound` → **exit 1**. The "3" in the issue text appears to conflate the variant name with a number; using 3 here would contradict the foundational exit-code contract. **Please confirm.**

## Engine (src/kv.rs)

- `add_key_def(key, KeyDef)` — general definition path (all types), persists via `save_schema`.
- `drop_key(key)` — full removal, rename_key transaction pattern.
- `update_key_meta(...)` — safe-only metadata edits, schema-only persist with rollback.
- `key_has_content(key)` — drives the `--force` gate (distinguishes empty/default vs. populated).
- `add_key_to_schema` retained unchanged for `push --create` (append-based, comment-preserving).

## Docs (all 6 surfaces)

`docs/src/kv.typ` (new Schema management section + deprecation notes), `architecture.typ` (schema-mutation model), `getting-started.typ`, `README.md`, regenerated HTML via `docs/build.sh`, and the tracked LLM bundle `docs/out/llm/mx-docs.md`.

## Test evidence

`cargo test --bin mx`: **1128 passed; 0 failed; 3 ignored**.

New coverage: `add` each of the five types + duplicate/invalid-name; `drop` empty/non-empty/unregistered/with-memory-link; `update` safe-metadata/clear-description/add-field/duplicate-field/unregistered; `key_has_content` empty-vs-populated; `parse_data_field_spec` all types + rejections. Rename delegation is exercised via the shared `rename_key` engine tests.

Live verification (scratch `MX_KV_SCHEMA`/`MX_KV_DATA`):
- `kv keys` vs `kv schema list` stdout: **byte-identical** (`cmp -s` clean); deprecation notice on stderr only.
- `drop` non-empty without `--force` → exit 4; with `--force` → exit 0; unregistered → **exit 1**.
- `push --create` still works through the shared path.
- `--type` on update → exit 4 with `kv migrate` pointer.
- Negative counter bounds (`--min -10`) parse (added `allow_hyphen_values`).

## Notes for reviewers

- HTML under `docs/out/html/` is gitignored; only `docs/out/llm/mx-docs.md` is tracked. Both regenerated.
- The repo working tree has pre-existing untracked junk (`C:/`, `examples/`, `memory-export/`). The `examples/win_lock_repro.rs` junk fails `cargo fmt --check` independently of this PR — it was moved aside only for the push hook and is not part of this branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,13 @@ mx kv set context --json '{"goal":"done","phase":"writing"}'
 mx kv set mytensor --json '[0.4, 0.6, 0.5]'
 echo '{"goal":"done"}' | mx kv set context --json -
 
-# Auto-create a key in the schema and push in one step
+# Schema management: define, list, edit, and remove keys
+mx kv schema add puns --type history --max-entries 500
+mx kv schema list                              # replaces 'kv keys'
+mx kv schema update puns --description "the good ones"
+mx kv schema drop puns --force                 # removes definition + data
+
+# Auto-create a key in the schema and push in one step (inline shortcut)
 mx kv push puns "the joke" --create history
 mx kv push ideas "wild thought" --create list --max-entries 500
 
@@ -182,7 +188,8 @@ mx kv migrate projects --dry-run
 mx kv migrate projects --prune
 
 # Rename a key (preserves all entries, IDs, and data)
-mx kv rename old_decisions archived_decisions
+mx kv schema update old_decisions --name archived_decisions
+# (the top-level 'mx kv rename' still works but is deprecated)
 
 # Per-entry memory links (bridge KV entries to the knowledge graph)
 mx kv push decisions "adopted memory links" --memory kn-abc123
--- a/docs/out/llm/mx-docs.md
+++ b/docs/out/llm/mx-docs.md
@@ -2878,14 +2878,19 @@ Schema fields:
 
 Keys can be added to the schema on the fly with `mx kv push --create`.
 When a key does not exist in the schema, `--create history` or
-`--create list` appends a new `[keys.<name>]` block to the TOML file and
-reloads the in-memory schema. This avoids manual schema editing for
-simple cases. See push for details and validation rules.
+`--create list` defines it and reloads the in-memory schema. This avoids
+manual schema editing for simple cases. See push for details and
+validation rules.
 
 For defining keys of **any** type up front (not just `history`/`list`),
 and for removing or editing existing definitions, use the `mx kv schema`
-subcommands. `push --create` is an inline shortcut that routes through
-the same engine path as `schema add`.
+subcommands. `push --create` is an inline shortcut whose user-facing
+surface stays limited to `history`/`list`, but it routes through the
+**same** engine path as `schema add` (the shared `add_key_def` method).
+One consequence: like every `schema` write, the create rewrites the
+schema file by round-tripping the TOML, which drops comments and custom
+formatting. If you hand-annotate your schema, expect those annotations
+to be lost when a key is created via `--create`.
 
 ### Agent keying
 
@@ -3260,8 +3265,10 @@ requires `--create` and has no effect if the key already exists.
 
 Key names are validated on creation: alphanumeric characters,
 underscores, and hyphens only, maximum 128 characters. Dots are rejected
-because they conflict with TOML key quoting. The new key block is
-appended to the schema file without reformatting existing content.
+because they conflict with TOML key quoting. Creation shares the
+`schema add` engine path, so it persists by round-tripping the schema
+TOML -- comments and custom formatting in the schema file are dropped,
+the same as any other `schema` write.
 
 ### Flags
 
@@ -4224,21 +4231,28 @@ The key name is validated the same way as elsewhere: alphanumeric,
 underscores, and hyphens only; maximum 128 characters; no dots. Defining
 a key that already exists is an error.
 
+Options are **type-appropriate**: passing an option that does not apply
+to the declared `--type` is rejected (exit code 4) rather than silently
+persisted. The valid pairings are: `--max-entries` (history, list);
+`--default` (counter, string, state); `--min` / `--max` (counter);
+`--fields` (state); `--data` (state, list). `--description` applies to
+every type.
+
 Only the schema file is written -- no data entry is created until the
 key is first written to.
 
 ### Flags
 
   **Flag**                          **Type**   **Description**
-  --------------------------------- ---------- --------------------------------------------------------------------------------------------------------------
+  --------------------------------- ---------- ----------------------------------------------------------------------------------------------------------------------------------
   `--type <type>`                   enum       Required. One of `counter`, `history`, `state`, `string`, `list`.
   `--max-entries <n>`               int        Cap for history/list keys; oldest entries drop when exceeded.
-  `--default <v>`                   str        Initial value for counter/string keys.
+  `--default <v>`                   str        Initial value for counter/string/state keys.
   `--min <n>`                       int        Minimum value for counters (clamped). Accepts negatives.
   `--max <n>`                       int        Maximum value for counters (clamped). Accepts negatives.
-  `--description <text>`            str        Human-readable description, shown by `schema list`.
+  `--description <text>`            str        Human-readable description, shown by `schema list`. Applies to all types.
   `--fields <a,b,c>`                list       Valid field names for a state key (comma-separated or repeatable).
-  `--data <name:type[:required]>`   str        Typed data-field definition (repeatable). `type` is one of `string`, `number`, `boolean`, `array`, `object`.
+  `--data <name:type[:required]>`   str        Typed data-field definition for state/list keys (repeatable). `type` is one of `string`, `number`, `boolean`, `array`, `object`.
 
 ### Examples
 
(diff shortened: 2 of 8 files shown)


coryzibell (commented):
## Verdictia Stern — review of PR #375 (`kv schema` suite, #363)

**Verdict: needs-changes** — one substantive contract gap, one inaccurate doc claim, one footgun, and a test gap on the exact safety property the issue called out. The engine transaction work is clean; the verbs behave; this is close.

### Critical

**C1. `push --create` does NOT route through the same path as `schema add` — violates AC and a doc claim.**
The acceptance criterion reads: *"`mx kv push --create` still works and routes through the same creation path as `schema add`."* It doesn't.
- `push --create` → `CreateType` (history/list only) → `add_key_to_schema()` (append-based, comment-preserving) — `src/handlers/kv.rs:683-689`, `src/kv.rs:1128`.
- `schema add` → `SchemaType` (all five) → `add_key_def()` (full TOML round-trip) — `src/handlers/kv.rs:1337`, `src/kv.rs:1197`.

Two distinct engine methods, two distinct enums, two distinct persistence behaviors. Verified live: `push --create counter` is rejected (`possible values: history, list`) while `schema add --type counter` works. Keeping the append path for `--create` to preserve comments is a *defensible* engineering call — but then the AC's "same path" wording needs an explicit Savorist waiver, because as written the criterion is unmet.

**C2. `docs/src/kv.typ` makes a false claim that contradicts `architecture.typ`.**
`kv.typ`: *"`push --create` is an inline shortcut that routes through the same engine path as `schema add`."* That is the C1 falsehood, stated as fact in user-facing docs. Meanwhile `architecture.typ` is honest — it documents `add_key_to_schema()` and `add_key_def()` as two separate methods. The two doc surfaces disagree. Fix `kv.typ` to match reality (and `architecture.typ`).

### Warnings

**W1. Type-inappropriate options are accepted silently — schema footgun.**
`add_key_def`/`update_key_meta` apply every option blindly with no type gate. Verified live: `schema update <counter> --add-field note:string` succeeds and writes a `[keys.ctr.data.note]` block onto a *counter*. Likewise `--min/--max` on a string, `--max-entries` on a counter, `--data` on non-state/list. Nonsensical schema persists without complaint. The issue specified these as *"type-appropriate options."* Recommend rejecting options that don't apply to the declared/existing type.

**W2. Rollback-on-failure path is entirely untested.**
The issue's acceptance criterion: *"`drop` and `update` follow the `rename_key` transaction pattern (data-first persist, rollback on failure)."* The rollback branches in `drop_key` (`kv.rs:1247-1253`) and `update_key_meta` (`kv.rs:1340-1343`) — the single most safety-critical logic in this PR — have **zero test coverage**. The 17 engine tests cover happy paths and validation errors only; none forces a write failure (e.g. read-only schema dir) to exercise the rollback. The reference `rename_key` rollback is also untested. "Follows the pattern" is asserted, not proven. Add at least one read-only-filesystem test per method that asserts in-memory state is restored after a failed persist.

### Suggestions / nits

**S1. No-op `schema update` rewrites the schema file.** `schema update ctr` with zero options prints "Updated" and re-serializes the TOML (dropping comments) for nothing. Verified live. Either short-circuit when no field is set, or warn. Given the documented comment-loss, a no-op that silently destroys annotations is unkind.

**S2. Exit code is internally consistent.** `drop` of an unregistered key → `KeyNotFound` → exit 1, matching `exit_code_for` (`handlers/kv.rs:16`), the canonical table (`kv.rs:99-102`), and every other verb. The issue prose's "exit 3" conflicts with the table; the implementer flagged it for the Savorist. The *code* is consistent with the convention it claims. No code change needed — this is a policy call for the Savorist, not a defect.

### Verified clean
- **Transaction pattern (drop/update):** `drop_key` faithfully mirrors `rename_key` — in-memory mutate, `save()` data-first with full rollback, then `save_schema()`. `update_key_meta` snapshots `original`, mutates, rolls back on `save_schema` failure and on duplicate-field. No partial-write window beyond the same narrow schema-write-after-data-write gap that already exists in `rename_key` and is documented identically in both. Sound.
- **Byte-identical `kv keys` vs `schema list`:** both now call the single shared `schema_list_handler` with identical format strings (`{:30} {:10} {}` / `{:30} {:10}`). Genuinely identical in code, not just in the PR body. Confirmed.
- **Deprecation notices:** `kv keys` and `kv rename` print the stderr notice; stdout unchanged; clap doc-comments mark both deprecated. Verified live (`keys` notice on stderr, listing on stdout).
- **Retype forbidden:** `schema update --type` errors, points at `kv migrate`, exit 4, no override. Verified live.
- **`--force` gate:** non-empty drop without `--force` errors (exit 4); empty/never-written drop silent. `key_has_content` correctly treats empty string / empty history-list / zero counter / unpopulated state as no-content. Verified live.
- **`allow_hyphen_values`** correctly set on `--min`/`--max` (both add and update) so negative bounds parse.
- **`parse_data_field_spec`** rejects bad type, missing type, empty name; accepts required/optional; tested.
- **Build + tests:** `cargo build` clean; `cargo test --bin mx kv` → 372 passed, 0 failed.

coryzibell (commented):
## Verdictia Stern — second-round verdict: APPROVE (one follow-up nit)

*(Posting as a comment — GitHub blocks self-approval on this account. Treat this as an approving verdict.)*

Read the fix commit three times. Every finding from round one is **genuinely** fixed — not paper-fixed. Verified at the code level, the test level, and live against the built binary. The nut is good.

### Per-finding verification

**C1 — push --create routes through add_key_def. FIXED.**
`handlers/kv.rs:707` — `push --create` now builds a `KeyDef` and calls `store.add_key_def(&key, def)`, the *literal same engine method* as `schema add` (`:1426`). The AC's "same creation path" is now true in code, not just in prose. The history/list surface limit is preserved by mapping `CreateType → ValueType` before constructing the def, and live `push --create counter` is still rejected at the clap layer (exit 2) — surface intact. The append-based path is correctly dropped.

**C2 — docs reconciled + bundle regenerated. FIXED.**
`kv.typ` no longer claims an append path; it states the unified `add_key_def` path and explicitly documents the comment-drop on `--create`. `architecture.typ` agrees, and honestly discloses the retained legacy helper. The regenerated LLM bundle (`docs/out/llm/mx-docs.md`) carries the same claims (lines 2887, 7830–7851) with no stale "appends a block for push --create" text remaining. No surface disagrees with another or with the post-C1 code.

**W1 — type-option gate. FIXED.**
`validate_type_options` gates both add (before persist, `handlers/kv.rs:1379`) and update (against the *existing* type, `:1498`), exiting 4 before any write. Table coverage is complete and matches the issue: max-entries↔history/list, default↔counter/string/state, min/max↔counter, fields↔state, data/add-field↔state/list, description↔all. 14 new tests assert behavior (6 pure-gate + 4 add/update reject + accepts-appropriate). Live: every inapplicable option exits 4 with a clear message; nonsense is not persisted.

**W2 — rollback tests. FIXED — and proven, not asserted.**
The `unwritable_under_file` seam plants a regular file and returns a path *under* it, so `create_dir_all(parent)` fails deterministically (no permission-bit flakiness). All three tests (drop_key, update_key_meta, rename_key) force the real data/schema write-failure path, then assert BOTH in-memory rollback (def + data restored to exact prior values) AND on-disk untouched-ness via reload. This is the exact safety property the issue called out, now under test.

**S1 — no-op update. FIXED.**
No-op `schema update` short-circuits with exit 4 *before* any serialize. Test asserts byte-identical file content + unchanged mtime. Live confirmation: md5 identical and mtime unchanged after a rejected no-op; hand annotations survive byte-for-byte.

### Regression scan — clean
Build clean, clippy clean, `cargo test --bin mx kv` → **390 passed, 0 failed** (was 372; +18 net). Prior-clean items spot-checked and intact: rename data-preservation, drop empty/non-empty, byte-identical schema_list, deprecation notices, transaction safety. Fix touches exactly 5 files, no scope creep.

### Follow-up nit (non-blocking)
`add_key_to_schema` is now an **unwired** standalone engine helper — zero non-test call sites. Acceptable to retain in *this* PR: it's `pub`, documented honestly in both rustdoc and architecture.typ, and carries its own tests (no dead-code warning, no misleading wiring). But it is dead weight on the public surface. Recommend a follow-up issue to remove it (or demote to `#[cfg(test)]`) once nothing external depends on it.

The verdict is the verdict. Ship it.

— *the smallest nod*

PR #382, opened 2026-06-11
fix(session): replace dead agent-prefix filter with UUID shape check

gabaum10:
## Bug

`find_most_recent_session()` in `src/session.rs` contained a dead filter:

```rust
if let Some(name) = file_path.file_name().and_then(|n| n.to_str())
    && name.starts_with("agent-")
{
    continue;
}
```

This guard was intended to skip sub-agent JSONL files. It never fired. Claude writes sub-agent transcripts into a dedicated subdirectory (`<project>/<session_id>/subagents/agent-*.jsonl`), not alongside root sessions in the flat `<project>/` directory. Every JSONL in each project root was included regardless of shape, making `find_most_recent_session()`'s default selection unreliable.

**Evidence:** During W420, the codex archive lost 5 sessions because this selector returned the wrong session. A pocket-skill workaround was added at the time to compensate.

**Scope:** This fix covers `find_most_recent_session()` only, which is the path used by `codex::archive` for default session resolution. The `--all` selector uses a separate code path and is not affected by this change.

## Fix

Replace the dead prefix guard with `is_uuid_shaped(stem)`, which is already defined in this file and is the correct discriminant:

- All real root sessions have UUID stems (full 36-char or short 8-char form).
- Stray files (scratch notes, partial writes, etc.) do not.
- Sub-agent files are unaffected — they live in a subdirectory and are found via `codex::archive::sources::find_agent_sessions`, not this walker.

Also updates the "no sessions found" error message to reflect the actual filter criterion.

## Known residual

`is_uuid_shaped` accepts any 8 hex-char string (see `session.rs:125`), so a file named `deadbeef.jsonl` in a project directory would pass the filter and could win the mtime sort over a real session. Same failure class as the original bug, but much narrower in practice (requires a stray file with an 8-char hex name to be more recently modified than any real session). Noting it so the reviewer sees we know.

## Test

Added `find_most_recent_session_uuid_filter_skips_non_uuid_files`, which:

- Proves `agent-` prefixed names do not pass the UUID filter (demonstrating the old guard was wrong).
- Proves bare full-UUID and short-UUID stems do pass.
- Proves non-session file names (`notes`, `README`, etc.) are correctly skipped.

```
cargo test session
running 39 tests
... all 39 passed, 0 failed
```

`cargo clippy --all-targets --all-features -- -D warnings` clean (verified by pre-push hook).

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -211,9 +211,24 @@ pub(crate) fn find_most_recent_session() -> Result<PathBuf> {
                 continue;
             }
 
-            // Skip agent sub-sessions — the legacy heuristic.
-            if let Some(name) = file_path.file_name().and_then(|n| n.to_str())
-                && name.starts_with("agent-")
+            // Only count files whose stem is UUID-shaped (bare UUID root
+            // sessions). The legacy `agent-` prefix guard never fired in
+            // practice because Claude writes sub-agent JSONLs into a
+            // dedicated subdirectory (`<project>/<session_id>/subagents/
+            // agent-*.jsonl`), not alongside the root session files. That
+            // dead filter caused `find_most_recent_session` to silently
+            // include every JSONL in the project root regardless of shape,
+            // making the default selector and --all unreliable. The UUID
+            // shape check is the correct discriminant: any non-UUID file
+            // (scratch notes, partial writes, etc.) is correctly skipped,
+            // and every real session — bare-UUID or otherwise — is picked
+            // up as long as its stem parses as a UUID.
+            //
+            // Evidence: W420 codex archive lost 5 sessions because the
+            // pocket-skill workaround had to compensate for this selector
+            // returning the wrong session.
+            if let Some(stem) = file_path.file_stem().and_then(|s| s.to_str())
+                && !is_uuid_shaped(stem)
             {
                 continue;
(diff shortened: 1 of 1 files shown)


coryzibell (approved):
Approving on the strength of a full Tsunderground review (Diffi). Verdict: **MERGE WITH FOLLOW-UP**.

The `is_uuid_shaped` replacement is the right shape of fix — positive allowlist over a dead denylist, properly anchored (36-char full form + 8-char short form, dashes at fixed positions, hex-only), no panic surface. CI green across the whole matrix (Gate, Format, Clippy `-D warnings`, Test on ubuntu/macos/windows). The author named his own residual (the 8-hex-char `deadbeef.jsonl` edge), which is vanishingly narrow and strictly better than status quo.

Two follow-ups tracked in #383 (non-blocking, so merging as-is):
1. Port the same fix to the `--all` path (`save_all_sessions`, `write.rs:641-645`) — it still carries the identical dead filter + an unvalidated session_id derivation. Grep surfaced 3 more sibling sites to audit.
2. Correct the over-claiming `--all` comment this PR adds — `--all` is a separate code path and isn't affected.

One Warmth. 🦊🔥

PR #399, opened 2026-07-02
fix: durable memory writes no longer exit 1 when post-write embed/anchor fails

gabaum10:
## Symptom

`mx memory append` (and its siblings) reported exit 1 on writes that had already landed. Callers treating exit 1 as "the write failed" re-fired — and the re-fire created duplicates. Hit in real use: a backgrounded `append` reported failure on a write that was already durable, the caller retried, and the entry got written twice.

## Root cause

The durable write commits first, then post-write side effects run: `auto_embed(...)?` and `auto_anchor(...)?`. Those `?` operators propagate *transient* side-effect failures — model cold-load, a websocket blip — straight to a non-zero exit, **after** the entry is already durable. So a fully-successful, committed write surfaced as a failure because an optional follow-up step hiccupped.

Diagnosis was trace-level: walking iteration counts and the exact `?` operators in the post-write blocks of `src/handlers/memory.rs`.

## Fix

Post-write embed/anchor is now non-fatal: on failure it logs a stderr warning noting that the entry itself is durable, and exits 0. This mirrors the existing non-fatal backup pattern already in the codebase. Applied symmetrically across **Append / Add / Update / Edit / Prepend / Restore**. Genuine write failures — where the durable write itself fails — still exit non-zero.

## Validation

Adversarial repro against the patched build: corrupted the cached ONNX blob to force `TractProvider::new` to fail at the exact diagnosed line.

- Transient side-effect failure → exit 0 + stderr warning + durable write confirmed by a follow-up graph read.
- Genuine failure (append to a nonexistent ID) → still exits 1.

New test file `tests/write_side_effect_nonfatal.rs`, 5/5 passing. `cargo fmt` and `cargo clippy` clean.

## Known follow-up (not in this PR)

`--no-embed` / `--no-auto-anchor` emit their "(... skipped)" skip-notices via `println!` to **stdout**, ahead of the JSON payload under `--json` — which corrupts the "stdout is JSON" contract. Surfaced by the new tests (the `add_plain_no_embed` helper drops `--json` to work around it). Pre-existing, independent of this change. Left for a separate call on the broader "human-facing notices never touch stdout under `--json`" rule.

## Note

The local test environment couldn't run `tests/trigger_check.rs` (SurrealDB IAM permissions — environmental, not a code issue). CI should exercise it properly.

### DIFF
--- a/src/handlers/memory.rs
+++ b/src/handlers/memory.rs
@@ -610,8 +610,29 @@ fn dedup_gate(
 /// write. `Skipped` is an `Ok` variant, never `Err` -- an `Err` would abort
 /// the whole batch via `?` for what is, by design, a successful no-op.
 enum WriteOutcome {
-    Written(Box<knowledge::KnowledgeEntry>, DedupProceedReason),
-    Skipped { duplicate_of: String },
+    Written(
+        Box<knowledge::KnowledgeEntry>,
+        DedupProceedReason,
+        WriteSideEffects,
+    ),
+    Skipped {
+        duplicate_of: String,
+    },
+}
+
+/// Outcome of the post-write embed/anchor side-effect steps in [`add_one`]
+/// (Cory's B1, PR #399 re-review). Both steps are non-fatal by design: a
+/// transient failure there must not abort the write or flip the exit code
+/// (the entry is already durable and read-back verified by the time either
+/// runs) -- but it also must not vanish into a stderr-only warning with zero
+/// signal in `--json` output. `None` means the step either succeeded or was
+/// never attempted (skipped via --no-embed / --no-auto-anchor, a deliberate
+/// no-op, not a failure -- those are surfaced separately by the existing
+/// "(embed skipped)" / "(auto-anchor skipped)" notices).
+#[derive(Default)]
+struct WriteSideEffects {
+    embed_deferred: Option<String>,
+    anchor_deferred: Option<String>,
 }
 
 /// Resolved arguments for a single standard-mode `memory add` write.
@@ -776,19 +797,28 @@ fn add_one(
     // Non-fatal: the entry is already durable (upserted + read-back verified
     // above), so a transient embed failure here must not surface as a
     // process exit failure -- that would make callers retry and duplicate
-    // an entry that already landed.
+    // an entry that already landed. The failure is still captured (not just
+    // eprintln'd) so callers building a `--json` payload can surface the
+    // degraded state instead of it vanishing into stderr (B1, PR #399
+    // re-review).
+    let mut side_effects = WriteSideEffects::default();
     if embed {
-        let _ = auto_embed(&id, db)
-            .map_err(|e| eprintln!("Warning: post-write embed failed (entry durable): {e}"));
+        if let Err(e) = auto_embed(&id, db) {
+            eprintln!("Warning: post-write embed failed (entry durable): {e}");
+            side_effects.embed_deferred = Some(e.to_string());
+        }
     } else {
         println!("  (embed skipped)");
     }
 
     // Auto-generate anchors. Gated by --no-auto-anchor / MX_SKIP_WRITE_ANCHOR.
-    // Non-fatal for the same reason as the embed step above.
+    // Non-fatal for the same reason as the embed step above; failure capture
+    // mirrors it too.
     if write_anchor_enabled(no_auto_anchor) {
-        let _ = auto_anchor(&id, db, None)
-            .map_err(|e| eprintln!("Warning: post-write anchor failed (entry durable): {e}"));
+        if let Err(e) = auto_anchor(&id, db, None) {
+            eprintln!("Warning: post-write anchor failed (entry durable): {e}");
+            side_effects.anchor_deferred = Some(e.to_string());
+        }
     } else {
         println!("  (auto-anchor skipped)");
     }
@@ -806,7 +836,11 @@ fn add_one(
         );
     }
 
-    Ok(WriteOutcome::Written(Box::new(entry), proceed_reason))
+    Ok(WriteOutcome::Written(
+        Box::new(entry),
+        proceed_reason,
+        side_effects,
+    ))
 }
 
 pub(crate) fn handle_memory(cmd: MemoryCommands, verbose: bool) -> Result<()> {
@@ -1639,8 +1673,8 @@ pub(crate) fn handle_memory(cmd: MemoryCommands, verbose: bool) -> Result<()> {
 
             // Skip = success, not failure: an identical entry already exists
             // this session, so there is nothing to write and nothing wrong.
-            let (entry, dedup_reason) = match outcome {
-                WriteOutcome::Written(e, reason) => (e, reason),
+            let (entry, dedup_reason, side_effects) = match outcome {
+                WriteOutcome::Written(e, reason, side_effects) => (e, reason, side_effects),
                 WriteOutcome::Skipped { duplicate_of } => {
                     if json {
                         println!(
@@ -1710,6 +1744,19 @@ pub(crate) fn handle_memory(cmd: MemoryCommands, verbose: bool) -> Result<()> {
                 }
                 if let DedupProceedReason::LookupFailed { msg } = &dedup_reason {
                     payload["dedup_lookup_warning"] = serde_json::json!(msg);
+                }
+                // B1 (PR #399 re-review): fold post-write embed/anchor
+                // side-effect failures into the same payload shape as the
+                // dedup fields above -- present only when the step actually
+                // failed, absent on success or on a deliberate --no-embed /
+                // --no-auto-anchor skip. This is the entry's only signal
+                // surface in --json mode; the stderr warning above still
+                // fires unconditionally for plain-mode callers.
+                if let Some(msg) = &side_effects.embed_deferred {
+                    payload["embed_deferred"] = serde_json::json!(msg);
+                }
+                if let Some(msg) = &side_effects.anchor_deferred {
+                    payload["anchor_deferred"] = serde_json::json!(msg);
                 }
                 println!("{}", serde_json::to_string_pretty(&payload)?);
             } else {
@@ -3157,7 +3204,11 @@ pub(crate) fn handle_memory(cmd: MemoryCommands, verbose: bool) -> Result<()> {
                     );
 
                     match entry_result {
-                        Ok(WriteOutcome::Written(entry, reason)) => {
+                        // Batch always passes embed=false / no_auto_anchor=true
+                        // (embedding is hoisted below, anchoring deferred to the
+                        // nightly run), so `side_effects` is always empty here --
+                        // ignored, not dropped by accident.
+                        Ok(WriteOutcome::Written(entry, reason, _side_effects)) => {
                             println!("  [{}] Added entry: {} ({})", line_idx + 1, entry.id, title);
                             // Batch has no --json mode -- stderr is the only
                             // signal surface here. Mirrors the standard
@@ -5236,7 +5287,7 @@ mod dedup_gate_tests {
         )
         .unwrap();
         let first_id = match first {
-            WriteOutcome::Written(e, _) => e.id,
+            WriteOutcome::Written(e, _, _) => e.id,
             WriteOutcome::Skipped { .. } => panic!("first write must not be a duplicate"),
         };
 
@@ -5551,7 +5602,7 @@ mod dedup_gate_tests {
         .expect("a dedup lookup failure must fail OPEN, not abort the write with Err");
 
         match outcome {
-            WriteOutcome::Written(entry, reason) => {
+            WriteOutcome::Written(entry, reason, _side_effects) => {
                 assert!(
                     matches!(reason, DedupProceedReason::LookupFailed { .. }),
                     "the proceed reason must record that the gate failed open on a lookup error"
--- a/src/handlers/memory.rs
+++ b/src/handlers/memory.rs
@@ -772,15 +772,23 @@ fn add_one(
 
     // Auto-generate embedding. Gated by the caller-resolved `embed` flag
     // (which reflects both --no-embed and MX_SKIP_WRITE_EMBED).
+    //
+    // Non-fatal: the entry is already durable (upserted + read-back verified
+    // above), so a transient embed failure here must not surface as a
+    // process exit failure -- that would make callers retry and duplicate
+    // an entry that already landed.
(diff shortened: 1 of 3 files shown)


coryzibell (changes requested):
## VERDICT: REQUEST CHANGES

The core change is right — durable-then-best-effort is the correct shape, and genuine write failures still exit non-zero (verified: `upsert`/read-back still use `?`). Durability is sound: `add_one` read-back-verifies and `bail!`s before any side effect runs (`src/handlers/memory.rs:446,62`). But two things go back down the chute: partial-success is invisible to the consumers that matter, and half the change (anchor) is entirely untested.

---

## Findings (ranked)

### Critical

**1. Partial-success is invisible to `--json` and exit-code consumers.** `src/handlers/memory.rs:479, 1126, 1815, 1898, 1992, 2082, 2212` (embed) and the anchor twins. On failure the entry lands durable-but-unsearchable (no embedding → absent from `--semantic` until a manual `embed --all`), and the *only* signal is a stderr `eprintln!`. Exit code is 0 — byte-identical to full success. Under `--json`, the payload gets no `embed_deferred`/`embedding_failed` field, so a structured caller (hearth/companions consume `--json`) reads clean success JSON and cannot detect the degraded state. The question "surfaced or fully silent" answers: *surfaced to a human on stderr, silent to every machine consumer.* Fix: add a structured field to the `--json` payload (e.g. `"embed_deferred": true`) so the degradation is programmatically detectable. This is the main lever.

**2. The anchor-failure path is entirely untested.** `tests/write_side_effect_nonfatal.rs`. All three failure tests use `mx_with_broken_model_cache`, which breaks *embed*. But `auto_embed` failing means the entry has no embedding, and `auto_anchor` early-returns `Ok(())` when `entry.embedding.is_none()` (`src/helpers.rs:405-407`). So the anchor error branch — 6 of the 12 changed call sites — never executes in any test. The PR claims symmetric Add/Update/Edit/Append/Prepend/Restore anchor coverage; zero anchor failures are actually forced. Add a test that forces `auto_anchor` to fail on an *already-embedded* entry.

### Warnings

**3. 4 of 6 single-entry sites have no non-fatal embed test.** Only Add (`:479`) and Append (`:1992`) are exercised. Update (`:1798`), Edit (`:1898`), Prepend (`:2082`), Restore (`:2212`) rely on "same code shape." Same shape, yes — but the commit message asserts all six, and untested is untested.

**4. Merge/interaction conflict with PR #402 is near-certain.** Both PRs edit `add_one`, the Add `--json` payload construction, and the `add-batch` embed block (#399 rewrites the hoisted embed pass at `:2700`; #402 adds a pre-write dedup gate to the same funnels). Semantically complementary — #402 dedups *before* the write, #399 softens side effects *after* — but the textual conflict in `add_one` and the batch block needs hand-reconciliation. More important: **the merged `--json` schema needs joint review.** #402 adds `skipped`/`duplicate_of`/`status`/`dedup_lookup_warning`/`dedup: bypassed_no_session`; #399 adds *none* for its degraded path (finding 1). See the Merge order section below.

**5. Reconcile is real but manual and unsignalled in the DB.** Confirmed `mx memory embed --all` re-embeds via `list_all` (`:2690-2734`), so a missed entry *will* be picked up — the PR comment's claim is accurate. But nothing triggers it automatically, it re-embeds the entire graph (not just the gaps), and there's no persistent "needs-embed" marker. A backgrounded caller that discards stderr leaves the entry silently unsearchable until a human remembers. Finding 1's structured field would also give a future targeted reconcile something to key on.

### Suggestions

**6. The `let _ = expr.map_err(|e| eprintln!(...))` idiom reads oddly** (all 12 sites). Using `map_err` for a side effect and then discarding a `Result<(),()>` obscures intent. `if let Err(e) = expr { eprintln!(...) }` says exactly what it does. Readability — the code is read a hundred times.

**7. The warning omits the entry id.** Every site prints `"post-write embed failed (entry durable): {e}"` with no `kn-` id. For the "reconcile later" story to be actionable, a human needs to know *which* entry. Add `&id`.

**8. Does not touch #398, correctly out of scope — but note the direction.** The diff wraps the existing `auto_anchor(&id, db.as_ref(), removed)?` (`:1817`) without altering anchor-merge plumbing, so #398 (`--add-anchor` clobbered by the auto-anchor pass) is neither fixed nor regressed. Caveat: making the anchor pass non-fatal makes an anchor-pass failure *quieter*, mildly at odds with #398's "at minimum fail loudly" ask on that path. Worth a line in the PR body.

**9. Tests are network- and environment-dependent** (self-documented): they require huggingface.co reachability, and `trigger_check.rs` couldn't run locally (SurrealDB IAM). CI must actually exercise both.

---

## Merge order

Agreed cross-PR plan for the three concurrent memory PRs, to prevent conflicts:

- **Recommended landing order: #401 → #402 → #399** (this PR lands **last**).
- **Rationale:** #402 changes `add_one`'s signature and return type (`Result<KnowledgeEntry>` → `Result<WriteOutcome>`) and adds `skipped`/`duplicate_of`/`status` to the `--json` payload. This PR should rebase on top of #402 and add its structured degraded-state field (finding 1, e.g. `embed_deferred`) into that already-richer JSON schema, so the JSON contract is designed **once, coherently**.
- **Semantics compose correctly:** #402's dedup-skip early-return means no post-write side effects run on a skip — which is the right behavior. #402 dedups before the write; #399 softens side effects after. No semantic conflict, only textual — resolved by landing in this order.

---

*The nut goes back to the examination table. Fix 1 and 2, fold the JSON field into #402's schema after rebase, and it earns the chute to the left.*

coryzibell (changes requested):
The core shape is right, and that's worth stating first: durable-then-best-effort is the correct design. The write is verifiably durable before any side effect runs — `add_one` upserts, read-back verifies and `bail!`s, and only then reaches embed and anchor — and genuine write failures still exit non-zero. On the Add path the failure is captured, printed, and returned rather than swallowed. `FailingScoredStore` and `anchor_failure_on_embedded_entry_propagates_err` close the Jul 4 anchor-test gap properly: it forces the error past the `embedding.is_none()` early return, which was the exact gap in question.

Three things blocking.

## 1. The new test file writes to the live graph

`tests/write_side_effect_nonfatal.rs:47-54` (`mx`) and `:72-79` (`mx_with_broken_model_cache`) set only:

```rust
.env("MX_HOME", env.dir.path())
.env("MX_CURRENT_AGENT", "test")
```

No `MX_SURREAL_MODE=embedded`, no `env_remove` of `MX_SURREAL_URL` / `NS` / `DB` / `USER` / `AUTH_LEVEL` / `ROOT`. In a shell that exports the network vars — which kautau's dev shell does — `cargo test --test write_side_effect_nonfatal` runs real `mx memory add` and `mx memory add-batch` invocations against the live knowledge graph. These tests write entries.

`main` now has an `isolate()` helper at `tests/common/mod.rs` from PR #420, merged today. Failing that, the equivalent already exists inside this stack at `tests/dedup_write_boundary.rs:48` — your own base branch. Both helpers in the new file need it.

## 2. `embed_deferred` landed on one of six write paths, and nothing asserts it

The field is defined at `src/handlers/memory.rs:625-632`, captured at `:807` (embed) and `:819` (anchor), emitted at `:1755-1759`. All on Add. The other five write paths accept `--json`, emit a payload, and report nothing:

| path | failure sites | JSON payload | structured signal |
|---|---|---|---|
| Update | `:2291`, `:2313` | `:2320` | none |
| Edit | `:2377`, `:2391` | `:2397` | none |
| Append | `:2471`, `:2485` | `:2491` | none |
| Prepend | `:2561`, `:2575` | `:2581` | none |
| Restore | `:2691`, `:2704` | `:2710` | none |

`mx memory update --json` emits `{"id": ..., "changes": [...]}` and nothing else. A structured caller reads clean success on a write whose embedding silently failed, and the entry stays absent from `--semantic` until someone remembers to run `embed --all`.

And nothing asserts the new field anywhere. Grep for `embed_deferred` / `anchor_deferred` returns the definition and emission lines plus one doc-comment, and no test. The three embed-failure integration tests assert exit 0 and that content persisted — both of which were true before this PR's JSON change. **Delete the emission lines and the whole suite stays green.** The one deliverable named the main lever has no safety net.

Thread the field into Update, Edit, Append, Prepend and Restore, and add a test that asserts it appears in `--json` on a failed embed. Or state on the record that Add is the only path with the guarantee and narrow the PR's claim to match — but pick one.

## 3. The thirteen warnings still omit the entry id

All thirteen sites print:

```
Warning: post-write embed failed (entry durable): {e}
```

No `kn-` id anywhere. Asked on Jul 4 and again on Jul 8. For a story whose whole premise is "it's durable, reconcile it later," the id is the one thing the human needs and the one thing not printed.

## Housekeeping

This is stacked on #402 (base is `feat/store-boundary-dedup`, and the merge-base is genuinely #402's head, so the phantom conflict with #401 is gone). Rebase onto `main` once #402 lands. Landing order stays #401 -> #402 -> #399.

Also worth knowing: there are no CI checks reported on this branch at all. `cargo check --tests` and `cargo clippy --all-targets` both run clean locally with the environment cleared, but nothing has ever run its tests automatically.

gabaum10:
Pushed `67b7ab2`. All three points addressed; one gap disclosed rather than closed, with the reason.

**Test isolation.** Both `mx()` and `mx_with_broken_model_cache()` now route through a local `isolate()` (`tests/write_side_effect_nonfatal.rs:59`) setting `MX_SURREAL_MODE=embedded` and `env_remove`ing the six network vars. Mirrors `tests/dedup_write_boundary.rs:48-58`. Kept local rather than factored into `tests/common/mod.rs` — that file doesn't exist at this branch point (it lands via #420 on main), and introducing it here would hand the eventual rebase a same-file conflict.

**Deferred-failure surfacing — threaded, not narrowed.** `embed_deferred`/`anchor_deferred` now capture and fold into the `--json` payload on Update, Edit, Append, Prepend and Restore, same shape as Add. Each `let _ = ....map_err(...)` became `if let Err(e)`, which also closes the idiom note from July.

**Entry id.** All thirteen sites now read `... (entry durable, id={id}): {e}`.

**On your falsification test.** You were right, and it was wider than the anchor field. At the commit you reviewed, **11 of the 12 emission lines were deletable with the suite green** — only Append's `embed_deferred` was load-bearing. Add's own wasn't: that test parsed `--json` and asserted only `v["id"]`.

Now all six `payload["embed_deferred"]` lines (`:1756 :2333 :2423 :2529 :2631 :2777`) have been commented out individually, verified to fail their mapped test, and restored. 0 of 6 deletable.

**The gap I did not close, and why.** The six `payload["anchor_deferred"]` lines remain deletable with the suite green. `auto_anchor` returns `Ok(())` when `entry.embedding.is_none()` (`src/helpers.rs:498`) before it touches the store, and the harness's only fault injector breaks the model cache — which guarantees no embedding, which guarantees that short-circuit. `auto_anchor` fails only on store calls. So no CLI-level test can reach an anchor-store failure regardless of injection seams, and `handle_memory` builds its own store per match arm with no seam to inject through; reaching those payload lines needs `handle_memory` to accept a store, which is its own change.

What I did instead is pin the capture one layer down: `FailingAnchorStore` plus `anchor_failure_is_captured_as_anchor_deferred_not_swallowed` drive `add_one` directly and assert `anchor_deferred.is_some()`. Commenting `side_effects.anchor_deferred = Some(...)` (`src/handlers/memory.rs:820`) fails it. That's parity with what `embed_deferred` had, not full coverage — I'd rather name it than have you find it.

**CHANGELOG.** Entry added under Unreleased/Fixed, naming the exit-status change as the load-bearing one plus the two paths where none of it surfaces (`add-batch` has no `--json` mode; `add --type` ignores it).

**Not rebased.** Your instruction was conditioned on #402 landing, and it hasn't. Rebasing now would detach this from its declared base. Landing order #401 → #402 → #399 stands.

**Two things for you, unchanged here.**

- Nine comments in `memory.rs` (`:1524 :2284 :2302 :2382 :2400 :2488 :2506 :2590 :2608`) say these steps run "if in network SurrealDB mode." They're false — `write_anchor_enabled` reads only `MX_SKIP_WRITE_ANCHOR`/`no_auto_anchor`, and neither `auto_embed` nor `auto_anchor` consults mode. They cost this round a wrong hypothesis before I checked the source. Your repo, your call.
- `FailingAnchorStore` is the fourth hand-written `KnowledgeStore` pass-through, after `CountingStore`, `FailingStore` and `FailingScoredStore`. Every new trait method now needs hand-adding in four places.

`cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features --locked -- -D warnings` and `cargo test --locked` all exit 0 — 1278 passed across 8 binaries.

coryzibell (approved):
Approving. Verified all of it by mutation, including the part you disclosed.

Deleting all six `embed_deferred` emissions reddens six tests, one per path. Deleting all six `anchor_deferred` emissions leaves the whole suite green, 1198 unit plus every integration binary. Killing the capture at `memory.rs:820` reddens your unit test. So the capture is pinned and the emission layer is not, exactly as you wrote it, and your reason holds: `auto_anchor` returns early with no embedding and the only injector guarantees no embedding.

Not blocking on that. Closing it needs `handle_memory` to take a store, which is its own change, and you named the gap instead of leaving it for me to find. Filed as #431.

Your nine comments are false as reported. Both gates read only the flag and the environment variable. They are all present verbatim on `main`, so this PR neither introduced nor worsened them, and reporting them against the repo instead of silently editing them inside an unrelated PR was the right call. Filed as #432. They cost you a wrong hypothesis for a round, which is exactly the price of a lying comment.

**On CI, read this before you go looking for a flake.** There isn't one. `wake-up.yml` fires on `pull_request: branches: [main]`, and this targets `feat/store-boundary-dedup`. CI has never run here and cannot run here while the base is #402's branch. That is structural, not intermittent. So this approval does not rest on a green check mark, and I am not going to pretend otherwise: it rests on the isolated run, 1198 green with the mutation test biting, plus 1332 across every binary on the rebased tree here. Retargeting to `main` after #402 lands is what turns CI on.

**What I did to this branch.** Rebased it onto the rebased #402 head and force-pushed. Zero conflicts. One source edit on top, same shape as the one on #402: your test doubles in `handlers/memory.rs` and `helpers.rs` implement `list_with_triggers`, which `main` removed from the `KnowledgeStore` trait, so I dropped the two delegators. Top commit, droppable. Base left pointing at #402's branch.

One suggestion for whenever you next touch these files: `isolate()` is now a third copy of the same helper. `mod common; common::isolate` would collapse it.

PR #401, opened 2026-07-03
feat: add --exclude-tags to memory search and list

gabaum10:
## Summary

Adds `--exclude-tags <PREFIXES>` (comma-separated, prefix-match) to `memory search` (both keyword and semantic paths) and `memory list`. Parse semantics match wake-fetch parity, so the same prefix syntax you already use for wake exclusion works here.

Motivation: the hearth just grew an archive tier (`tier/archived`), and archived entries should be excluded from retrieval by default. This flag is the retrieval-side lever for that — pull the archive back in only when you ask for it.

## Design notes

Semantic exclusion is pushed down to the **candidate-set level**, not applied as an after-the-fact filter on an already-limited result set:

- **Unchunked phase** filters in the SQL `WHERE` via bound params *before* `LIMIT`, so exclusion narrows the candidate set the database ranks, not the tail.
- **Embedding-chunk phase** has no tag edge to join against, so it uses a **bounded iterative over-fetch loop**: it grows the fetch window and re-scans until the limit is filled or the scan budget is exhausted. This guarantees an all-archived top window can't silently thin results below the requested limit.

The DB-layer `KnowledgeFilter` carries a pre-parsed `exclude_tag_prefixes: Vec<String>` (parsed once via `parse_exclude_prefixes` in `helpers.rs`); the CLI-facing `EntryFilter::exclude_tags` stays a raw CSV `Option<String>` to keep exact wake-fetch parse parity.

## Two honest behavior notes

1. **No-flag chunk path now grows its fetch window to fill the limit.** Previously the chunk path used a fixed `limit * 3` window. It now over-fetches iteratively to fill the requested limit. This is a strict improvement (fewer silently-short result sets), but it is **not byte-identical** to the old fixed-window behavior — a no-flag query may now return chunks the old fixed window would have dropped.

2. **Capacity characteristic under archive-dominated rankings.** When archived chunks dominate the top rankings, an excluded semantic query can run up to ~5 bounded scans instead of 1 to fill the limit. It's bounded, so no runaway — but it's more DB work in that specific case. The future fix, if it ever matters in practice, is to denormalize a tier flag directly onto `embedding_chunk` so the chunk phase can filter in SQL like the unchunked phase does.

## Sibling note

A separate store-boundary dedup PR (W447-mx-dedup) follows this one. Landing order is your call — this PR is deliberately minimal, scoped only to retrieval exclusion, to unblock the archive tier without dragging the dedup change along with it.

## Docs regeneration note

`docs/out/llm/mx-docs.md` was regenerated (`docs/build.sh`) as part of this change, which pulls in the already-merged `MX_SURREAL_*` env-var section move from `getting-started.typ` — main's committed copy of the generated file was stale relative to that prior merge. That's why the diff on this file is larger than the `--exclude-tags` additions alone; it's unrelated churn made correct, not unrelated churn left unexplained.

## Test plan

- 1166/1166 unit tests pass (includes new coverage in `src/surreal_db/tests.rs` for both exclusion phases and the over-fetch loop)
- `cargo fmt -- --check` clean
- Manual verification of keyword, semantic, and list exclusion paths against the archive tier

### DIFF
--- a/docs/out/llm/mx-docs.md
+++ b/docs/out/llm/mx-docs.md
@@ -1330,7 +1330,7 @@ Several read commands (`search`, `list`) share a common set of filter
 flags. These are documented once here and referenced below.
 
   **Flag**                     **Type**   **Description**
-  ---------------------------- ---------- -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+  ---------------------------- ---------- --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
   `-c, --category`             `string`   Filter by category (comma-separated).
   `--json`                     `flag`     Output as JSON.
   `--mine`                     `flag`     Show only your private entries.
@@ -1345,7 +1345,7 @@ flags. These are documented once here and referenced below.
   `--missing-resonance-type`   `flag`     Filter to entries WITHOUT a resonance type.
   `--limit`                    `int`      Limit number of results.
   `--tags`                     `string`   Filter by tags (comma-separated, matches any).
-  `--exclude-tags`             `string`   Comma-separated list of tag prefixes. Drops any entry that carries at least one tag prefix-matching at least one of these values -- useful for excluding whole tag namespaces (e.g. `tier/` removes every entry tagged `tier/<anything>`). Empty segments from trailing commas are ignored. Same parsing and prefix-match semantics as `wake-fetch`'s `--exclude-tags`.
+  `--exclude-tags`             `string`   Comma-separated tag prefixes to drop (e.g. `tier/` excludes every entry tagged `tier/<anything>`). Same parsing and prefix-match semantics as `wake-fetch`'s `--exclude-tags`.
 
 ::: {.admonition .note}
 **NOTE:** **Visibility default:** `list` and `search` run
@@ -1383,6 +1383,13 @@ therefore admit an entry under `wake` but exclude it under
 `transformative` resonance types are decay-exempt, so they filter
 identically under both. This divergence is intentional for now; an
 explicit `--resonance-basis raw|decayed` flag is tracked in #404.
+:::
+
+::: {.admonition .note}
+**NOTE:** `--exclude-tags` takes a comma-separated list of tag prefixes.
+Any entry carrying at least one tag that prefix-matches at least one
+supplied value is dropped -- useful for excluding whole tag namespaces.
+Empty segments from trailing commas are ignored.
 :::
 
 ## `mx memory show`
@@ -1485,10 +1492,17 @@ to have embeddings generated via `mx memory embed`.
 :::
 
 ::: {.admonition .note}
-**NOTE:** With `--semantic`, `--exclude-tags` is applied inside the
-candidate-set query itself, not as a post-filter -- excluded entries
-never occupy a ranked slot, so `--limit` still returns a full page even
-when top-ranked neighbors carry an excluded tag.
+**NOTE:** With `--semantic`, `--exclude-tags` exclusion happens in two
+phases: unchunked entries are excluded inside the candidate-set query
+itself (SQL `WHERE`), while chunked entries are filtered on the Rust
+side after hydration, since `embedding_chunk` has no direct tag edge to
+filter on in SQL. Either way, excluded entries are dropped before the
+final sort and truncate, so `--limit` returns a full page as long as the
+search stays within its scan budget (`limit * 50`, minimum 1000 chunk
+candidates). If that budget is exhausted before enough non-excluded
+results are found -- which can happen when excluded tags dominate the
+ranked candidates -- the result set may come back short, and a warning
+is logged when that happens.
 :::
 
 ::: {.admonition .tip}
--- a/docs/out/llm/mx-docs.md
+++ b/docs/out/llm/mx-docs.md
@@ -1319,7 +1319,8 @@ include all per-entry fields (`category`, `title`, `content`,
 `source_agent`, `tags`, `private`, `resonance`, `type`, etc.) on each
 line. There are no batch-wide field overrides except `--no-embed`. This
 design supports heterogeneous batches (facts, person nodes, summaries,
-blooms) in a single invocation --- which is the primary pocket use-case.
+blooms) in a single invocation --- which is the primary batch-write
+use-case.
 :::
 
 ## Reading entries {#reading}
--- a/docs/out/llm/mx-docs.md
+++ b/docs/out/llm/mx-docs.md
@@ -1330,7 +1330,7 @@ Several read commands (`search`, `list`) share a common set of filter
 flags. These are documented once here and referenced below.
 
   **Flag**                     **Type**   **Description**
-  ---------------------------- ---------- -------------------------------------------------------------------------------------------------------------------------------------
+  ---------------------------- ---------- -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
   `-c, --category`             `string`   Filter by category (comma-separated).
   `--json`                     `flag`     Output as JSON.
   `--mine`                     `flag`     Show only your private entries.
@@ -1345,6 +1345,7 @@ flags. These are documented once here and referenced below.
   `--missing-resonance-type`   `flag`     Filter to entries WITHOUT a resonance type.
   `--limit`                    `int`      Limit number of results.
   `--tags`                     `string`   Filter by tags (comma-separated, matches any).
+  `--exclude-tags`             `string`   Comma-separated list of tag prefixes. Drops any entry that carries at least one tag prefix-matching at least one of these values -- useful for excluding whole tag namespaces (e.g. `tier/` removes every entry tagged `tier/<anything>`). Empty segments from trailing commas are ignored. Same parsing and prefix-match semantics as `wake-fetch`'s `--exclude-tags`.
 
 ::: {.admonition .note}
 **NOTE:** **Visibility default:** `list` and `search` run
@@ -1424,6 +1425,11 @@ mx memory list -c discovery,decree --min-resonance 5
 mx memory list --missing-wake-phrase --limit 20
 ```
 
+``` bash
+# Exclude a whole tag namespace
+mx memory list --category person --exclude-tags 'tier/'
+```
+
 ::: {.admonition .note}
 **NOTE:** `list` accepts all shared filter flags documented above,
 including the public-only visibility default and the hidden-private
@@ -1461,11 +1467,28 @@ mx memory search "agent bootstrap" -c recipe,method --limit 5
 mx memory search "retry pattern" --activate
 ```
 
+``` bash
+# Exclude a whole tag namespace from keyword search
+mx memory search "onboarding" --exclude-tags 'tier/'
+```
+
+``` bash
+# Exclude a whole tag namespace from semantic search
+mx memory search "onboarding" --semantic --exclude-tags 'tier/'
+```
+
 ::: {.admonition .note}
 **NOTE:** `search` accepts all shared filter flags, including the
 public-only visibility default and the hidden-private stderr hint (the
 hint is suppressed under `--semantic`). Semantic search requires entries
 to have embeddings generated via `mx memory embed`.
+:::
+
+::: {.admonition .note}
+**NOTE:** With `--semantic`, `--exclude-tags` is applied inside the
+candidate-set query itself, not as a post-filter -- excluded entries
+never occupy a ranked slot, so `--limit` still returns a full page even
+when top-ranked neighbors carry an excluded tag.
 :::
 
 ::: {.admonition .tip}
--- a/docs/out/llm/mx-docs.md
+++ b/docs/out/llm/mx-docs.md
@@ -1319,7 +1319,7 @@ include all per-entry fields (`category`, `title`, `content`,
 `source_agent`, `tags`, `private`, `resonance`, `type`, etc.) on each
 line. There are no batch-wide field overrides except `--no-embed`. This
 design supports heterogeneous batches (facts, person nodes, summaries,
-blooms) in a single invocation --- which is the primary batch-write
+blooms) in a single invocation --- which is the primary pocket
 use-case.
 :::
 
--- a/docs/out/llm/mx-docs.md
+++ b/docs/out/llm/mx-docs.md
@@ -1319,8 +1319,7 @@ include all per-entry fields (`category`, `title`, `content`,
 `source_agent`, `tags`, `private`, `resonance`, `type`, etc.) on each
 line. There are no batch-wide field overrides except `--no-embed`. This
 design supports heterogeneous batches (facts, person nodes, summaries,
-blooms) in a single invocation --- which is the primary pocket
-use-case.
+blooms) in a single invocation --- which is the primary pocket use-case.
 :::
 
 ## Reading entries {#reading}
@@ -1331,7 +1330,7 @@ Several read commands (`search`, `list`) share a common set of filter
 flags. These are documented once here and referenced below.
 
   **Flag**                     **Type**   **Description**
-  ---------------------------- ---------- --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+  ---------------------------- ---------- -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
   `-c, --category`             `string`   Filter by category (comma-separated).
   `--json`                     `flag`     Output as JSON.
   `--mine`                     `flag`     Show only your private entries.
@@ -1346,7 +1345,7 @@ flags. These are documented once here and referenced below.
   `--missing-resonance-type`   `flag`     Filter to entries WITHOUT a resonance type.
   `--limit`                    `int`      Limit number of results.
   `--tags`                     `string`   Filter by tags (comma-separated, matches any).
-  `--exclude-tags`             `string`   Comma-separated tag prefixes to drop (e.g. `tier/` excludes every entry tagged `tier/<anything>`). Same parsing and prefix-match semantics as `wake-fetch`'s `--exclude-tags`.
+  `--exclude-tags`             `string`   Comma-separated tag prefixes to drop (e.g. `tier/` excludes every entry tagged `tier/<anything>`). Specify once; comma-separate multiple prefixes -- repeating the flag is a parse error. Same parsing and prefix-match semantics as `wake-fetch`'s `--exclude-tags`.
 
 ::: {.admonition .note}
 **NOTE:** **Visibility default:** `list` and `search` run
@@ -1390,7 +1389,9 @@ explicit `--resonance-basis raw|decayed` flag is tracked in #404.
 **NOTE:** `--exclude-tags` takes a comma-separated list of tag prefixes.
 Any entry carrying at least one tag that prefix-matches at least one
 supplied value is dropped -- useful for excluding whole tag namespaces.
-Empty segments from trailing commas are ignored.
+Empty segments from trailing commas are ignored. Pass the flag once;
+repeating it (e.g. `--exclude-tags a --exclude-tags b`) is a parse
+error, not a silent override.
 :::
 
 ## `mx memory show`
--- a/docs/src/memory.typ
+++ b/docs/src/memory.typ
@@ -121,7 +121,7 @@ per-entry fields (`category`, `title`, `content`, `source_agent`, `tags`,
 `private`, `resonance`, `type`, etc.) on each line. There are no batch-wide
 field overrides except `--no-embed`. This design supports heterogeneous batches
 (facts, person nodes, summaries, blooms) in a single invocation --- which is
-the primary pocket use-case.]
+the primary batch-write use-case.]
 
 
 // ═══════════════════════════════════════════════════════════════════════
--- a/docs/src/memory.typ
+++ b/docs/src/memory.typ
@@ -152,7 +152,7 @@ flags. These are documented once here and referenced below.
   [`--missing-resonance-type`], [`flag`], [Filter to entries WITHOUT a resonance type.],
   [`--limit`],         [`int`],    [Limit number of results.],
   [`--tags`],          [`string`], [Filter by tags (comma-separated, matches any).],
-  [`--exclude-tags`],  [`string`], [Comma-separated tag prefixes to drop (e.g. `tier/` excludes every entry tagged `tier/<anything>`). Same parsing and prefix-match semantics as `wake-fetch`'s `--exclude-tags`.],
+  [`--exclude-tags`],  [`string`], [Comma-separated tag prefixes to drop (e.g. `tier/` excludes every entry tagged `tier/<anything>`). Specify once; comma-separate multiple prefixes -- repeating the flag is a parse error. Same parsing and prefix-match semantics as `wake-fetch`'s `--exclude-tags`.],
 )
 
 #note[*Visibility default:* `list` and `search` run *public-only* by default.
@@ -189,7 +189,8 @@ flag is tracked in \#404.]
 #note[`--exclude-tags` takes a comma-separated list of tag prefixes. Any entry
 carrying at least one tag that prefix-matches at least one supplied value is
 dropped -- useful for excluding whole tag namespaces. Empty segments from
-trailing commas are ignored.]
+trailing commas are ignored. Pass the flag once; repeating it (e.g.
+`--exclude-tags a --exclude-tags b`) is a parse error, not a silent override.]
 
 #command(
   "mx memory show",
--- a/docs/src/memory.typ
+++ b/docs/src/memory.typ
@@ -152,7 +152,7 @@ flags. These are documented once here and referenced below.
   [`--missing-resonance-type`], [`flag`], [Filter to entries WITHOUT a resonance type.],
   [`--limit`],         [`int`],    [Limit number of results.],
   [`--tags`],          [`string`], [Filter by tags (comma-separated, matches any).],
-  [`--exclude-tags`],  [`string`], [Comma-separated list of tag prefixes. Drops any entry that carries at least one tag prefix-matching at least one of these values -- useful for excluding whole tag namespaces (e.g. `tier/` removes every entry tagged `tier/<anything>`). Empty segments from trailing commas are ignored. Same parsing and prefix-match semantics as `wake-fetch`'s `--exclude-tags`.],
+  [`--exclude-tags`],  [`string`], [Comma-separated tag prefixes to drop (e.g. `tier/` excludes every entry tagged `tier/<anything>`). Same parsing and prefix-match semantics as `wake-fetch`'s `--exclude-tags`.],
 )
 
 #note[*Visibility default:* `list` and `search` run *public-only* by default.
@@ -185,6 +185,11 @@ threshold can therefore admit an entry under `wake` but exclude it under
 resonance types are decay-exempt, so they filter identically under both. This
 divergence is intentional for now; an explicit `--resonance-basis raw|decayed`
 flag is tracked in \#404.]
+
+#note[`--exclude-tags` takes a comma-separated list of tag prefixes. Any entry
+carrying at least one tag that prefix-matches at least one supplied value is
+dropped -- useful for excluding whole tag namespaces. Empty segments from
+trailing commas are ignored.]
 
 #command(
   "mx memory show",
@@ -238,10 +243,17 @@ visibility default and the hidden-private stderr hint (the hint is suppressed
 under `--semantic`). Semantic search requires entries to have embeddings
 generated via `mx memory embed`.]
 
-#note[With `--semantic`, `--exclude-tags` is applied inside the candidate-set
-query itself, not as a post-filter -- excluded entries never occupy a ranked
-slot, so `--limit` still returns a full page even when top-ranked neighbors
-carry an excluded tag.]
+#note[With `--semantic`, `--exclude-tags` exclusion happens in two phases:
+unchunked entries are excluded inside the candidate-set query itself (SQL
+`WHERE`), while chunked entries are filtered on the Rust side after
+hydration, since `embedding_chunk` has no direct tag edge to filter on in
+SQL. Either way, excluded entries are dropped before the final sort and
+truncate, so `--limit` returns a full page as long as the search stays
+within its scan budget (`limit * 50`, minimum 1000 chunk candidates). If
+that budget is exhausted before enough non-excluded results are found --
+which can happen when excluded tags dominate the ranked candidates -- the
+result set may come back short, and a warning is logged when that
+happens.]
 
 #tip[By default, search does not activate results -- browsing is not the same as
 engagement. Use `--activate` when you are intentionally consuming the results
--- a/docs/src/memory.typ
+++ b/docs/src/memory.typ
@@ -121,7 +121,7 @@ per-entry fields (`category`, `title`, `content`, `source_agent`, `tags`,
 `private`, `resonance`, `type`, etc.) on each line. There are no batch-wide
 field overrides except `--no-embed`. This design supports heterogeneous batches
 (facts, person nodes, summaries, blooms) in a single invocation --- which is
-the primary batch-write use-case.]
+the primary pocket use-case.]
 
 
 // ═══════════════════════════════════════════════════════════════════════
--- a/docs/src/memory.typ
+++ b/docs/src/memory.typ
@@ -152,6 +152,7 @@ flags. These are documented once here and referenced below.
   [`--missing-resonance-type`], [`flag`], [Filter to entries WITHOUT a resonance type.],
   [`--limit`],         [`int`],    [Limit number of results.],
   [`--tags`],          [`string`], [Filter by tags (comma-separated, matches any).],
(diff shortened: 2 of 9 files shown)


coryzibell (changes requested):
## Verdict: REQUEST CHANGES

The exclusion logic itself is correct, injection-safe, and genuinely well-tested. The changes requested are about scope discipline and one untested semantic gap — not the core algorithm, which holds up.

---

### Findings, ranked

**1. (Blocking) Repeated `--exclude-tags` flag silently keeps only the last value — untested, and inconsistent with `--tags`.**
`src/cli.rs:208-209` — `exclude_tags: Option<String>` with `#[arg(long, value_name = "PREFIXES")]`. No `value_delimiter`, no `append` action. The sibling `tags` field (`src/cli.rs:200-201`) is `Option<Vec<String>>` with `value_delimiter = ','`, so `--tags a --tags b` accumulates. But `--exclude-tags a --exclude-tags b` silently overrides to `b` only — clap's default `Set` action takes the last occurrence and does not error. A user who exclude-filters two namespaces across two flags loses the first with no signal. There is no test for this. The `Option<String>` shape is justified as deliberate wake-fetch parse parity (`src/cli.rs:204-207`), which is a fair argument for the CSV form — but the multi-occurrence behavior must be pinned with a test, and either rejected, documented, or made to accumulate.

**2. (Blocking) Drop the unrelated docs drift from the diff, or justify it.**
- `docs/out/llm/mx-docs.md:9-39` removes the `MX_SURREAL_MODE/URL/NS/DB` env-var table and replaces it with a pointer. That change has **no source counterpart in this diff** — its source (`docs/src/getting-started.typ:248`) already carries the new text from a *prior* change. This is a generated file (`docs/build.sh`) catching up to stale source, folded in by regeneration. Benign, but it muddies the diff and inflates the line count.
- `docs/src/memory.typ:130` "pocket use-case" → "batch-write use-case" is a genuine unrelated wording tweak.
Neither belongs in a `--exclude-tags` PR. Split them out or add a one-line note in the PR body explaining why they ride along.

**3. (Warning) A behavior change to the no-flag semantic path is bundled into a flag-addition PR.**
`src/surreal_db/knowledge.rs:840-1004` — the chunk phase's fixed `limit * 3` over-fetch was replaced with an iterative grow-to-fill loop (`chunk_fetch` grows to `max_chunk_fetch = limit*50 .min(1000)`). This fires for *every* semantic search, not just `--exclude-tags` ones. It is acknowledged (PR "behavior note 1") and is a defensible bugfix (the old fixed window silently thinned category/resonance-filtered results — see the new test at `src/surreal_db/tests.rs:1458`). But a no-flag `search --semantic` can now return chunks the old window dropped. It composes fine, it is guarded, and there is coverage — flagging that a general retrieval-shape change rides in under a "deliberately minimal" flag PR.

**4. (Suggestion) The over-fetch loop re-runs the full chunk scan from scratch each iteration.**
`src/surreal_db/knowledge.rs:850-864` — each pass issues a fresh `SELECT ... FROM embedding_chunk ... LIMIT $chunk_limit` with a doubled limit, re-scanning and re-scoring rows already seen. `resolved_chunk_entries` (line 845) correctly prevents re-*hydration* (the expensive `get_knowledge_async`), but the SQL scan itself repeats up to ~4-5 times in the archive-dominance case. Bounded, acknowledged in the PR, and the denormalized-tier-flag fix is already named as the real remedy. Acceptable as-is; noting for the record.

**5. (Relevance to #400) This diff does NOT touch the privacy/visibility path — correctly orthogonal.**
Both entry points still resolve context identically to before: `src/handlers/memory.rs:519` (search) and `:581` (list) call `resolve_agent_context(filter.mine, filter.include_private)`, unchanged. The new exclude clause is `AND`-appended *after* `build_visibility_filter(ctx)` in `src/surreal_db/knowledge.rs:942`, and the chunk phase still hydrates through `get_knowledge_async(entry_id, ctx)` (`:931`), so the `include_private = false` default that causes #400's undercount is untouched. This PR neither fixes nor regresses #400 — it is a clean, independent axis. Stating plainly so nobody assumes the archive-tier work addressed the private-entry omission; it did not.

---

### What's correct (verified, not assumed)

The author earned this section.

- **No injection.** `build_exclude_tags_filter` (`src/surreal_db/knowledge.rs:692-712`) interpolates only generated parameter *names* (`exclude_prefix_0…`); prefix *values* are always bound via `.bind()` (`:778-780`). Caller-controlled input never reaches the SQL text.
- **AND/OR precedence is right.** `AND array::len(->tagged_with->tag[WHERE cond0 OR cond1 …]) = 0` — the OR lives inside the self-contained graph-traversal filter, then the `= 0` (keep rows where no tag matches any prefix) is `AND`-joined to the main WHERE. No precedence ambiguity.
- **Tag in both include and exclude → exclusion wins,** and it is tested (`src/helpers.rs:606-626`, `exclusion_wins_over_overlapping_include_tag`).
- **Exclusion applied server-side for the unchunked phase (before LIMIT)**, so a limited result set is never thinned — the exact interaction the review worried about. The chunk phase, which cannot push to SQL, uses the grow-to-fill loop instead of a post-filter-on-limited-set. Both are covered end-to-end against a real in-memory DB (`src/surreal_db/tests.rs:1163`, `:1267`), including the top-ranked-archived-neighbors case.
- **Keyword/list stay exact via post-filter** precisely because those queries issue no DB-level LIMIT, and there is a landmine test guarding that contract (`src/surreal_db/tests.rs:1064`, `keyword_search_exclude_survives_past_hypothetical_limit`).
- **Consistent semantics across search and list**, shared `EntryFilter`, single `parse_exclude_prefixes`/`keep_after_exclude` definition (correctly consolidated into `src/helpers.rs:401-426`, de-duplicated from the handler), docs updated for both commands and both search modes.

---

### Merge order

Agreed cross-PR plan for the three concurrent memory PRs:

- **Recommended landing order: #401 → #402 → #399** (this PR lands **first**).
- **Rationale:** this PR is the most independent of the three. #402 and #399 both rewrite `add_one` and the add/batch write funnels in `src/handlers/memory.rs` and will conflict textually with each other; this PR only brushes `memory.rs`/`knowledge.rs` on the **read** path. Landing it first minimizes total rebase work.

---

**Summary:** Correct, injection-safe, thoroughly-tested `--exclude-tags` with a sound server-side/client-side split and no impact on #400's visibility path. Requesting changes for two things: pin/fix the repeated-flag silent override at `src/cli.rs:208` with a test, and drop (or justify) the unrelated docs drift.

coryzibell (changes requested):
Sixty days on and this still holds up as real work. Two things in it are worth saying before anything else, because they're the reason to finish this rather than shelve it.

## The good news, and it's substantial

**This PR fixes a recall bug that is live on `main` right now.** Main's fixed `limit * 3` window doesn't just return short pages — it returns *wrong* ones. It silently substitutes bottom-of-table entries for genuinely better-ranked ones. Measured against ground truth on the live graph (1538 entries, 328 semantically searchable):

| query | main | #401 | truly available |
|---|---|---|---|
| `--min-resonance 9 --limit 10` | 4 | 10 | 15 |
| `--min-resonance 8 --limit 50` | 31 | 50 | 77 |

And at `wake ritual --limit 300`: main omits 28 entries whose true ranks are 257-300, and pads the page with 19 entries ranked 305-328. That is a shipped bug in `main` that this PR repairs, and it's a stronger argument for merging than the one the PR body makes for itself.

**The batch hydration is the largest user-visible win in the diff, and nobody claimed it.** Mean wall time across 67 paired live queries: main 3988ms, #401 2738ms. #401 is flat in `limit` where main degrades linearly:

| limit | main | #401 |
|---|---|---|
| 20 | 4575ms | 2639ms |
| 100 | 8338ms | 2787ms |
| 500 | 12349ms | 3127ms |

4x faster at limit 500. That's `get_knowledge_batch_async` replacing an N+1, not the fill loop.

Both of those are keepers. Which is exactly why the next part has to get fixed rather than argued around.

## Blocker: the batch hydration queries never bind, so tags come back empty

`src/surreal_db/relationships.rs:296` (`get_tags_for_entries_async`) and `:378` (`get_applicability_for_entries_async`) both filter with:

```sql
WHERE in IN $knowledge
```

`in` is a SurrealQL keyword. In that position the clause does not bind the edge's in-field, so it matches nothing — and SurrealDB returns zero rows *without an error*. Then `response.take(0).unwrap_or_default()` at both call sites turns "the query matched nothing" into "this entry has no tags." The failure is completely silent.

Measured on an in-memory store, same fixture, same bound `Vec<Thing>`:

| clause | rows |
|---|---|
| `WHERE in IN $k` | 0 |
| `WHERE $k CONTAINS in` | 1 (correct) |
| `WHERE in INSIDE $k` | 0 |

Three consequences, all measured on the live graph:

**1. Every chunked entry comes back from semantic search with tags and applicability stripped.** No flags needed, visible in default human-readable output. On `constellation family identity --limit 100`, identical 100 ids in both binaries:

| entry class | tags preserved | tags lost |
|---|---|---|
| chunked (`chunk_count > 0`) | 1 | **89** |
| unchunked (`chunk_count = 0`) | 10 | 0 |

A clean split on chunk status. The `Tags:` line simply disappears from the output for chunked entries. `applicability` is emptied the same way. There is no re-hydration path downstream — `get_knowledge_batch_async` builds the entry via `record.into_knowledge_entry(tags, applicability)`, that entry is cloned into `scored_entries`, and the renderer prints `entry.tags` directly.

**2. `--exclude-tags` is a no-op on chunked entries.** Because `entry.tags` is empty, `keep_after_exclude(&[], prefixes)` is unconditionally true. Test on the live graph with `--exclude-tags august`:

| entry | chunk_count | carries `august` | outcome |
|---|---|---|---|
| kn-57b770c7 | 0 | yes | correctly dropped |
| kn-bd0b74cd | 0 | yes | correctly dropped |
| kn-eb59be4a | 4 | yes | **wrongly kept** |
| kn-08ebe655 | 3 | yes | **wrongly kept** |

Reproduced again with `--exclude-tags constellation`. Exclusion survives only on the unchunked path, where it's pushed into SQL as a graph traversal and never touches the broken helper. That's 73 of 328 searchable entries working, and **255 of 328 — 78% — failing, and failing OPEN**: it returns what you asked it to hide. The archive tier is long-form, therefore chunked, therefore precisely the population the feature was written for.

**3. `semantic_search_excludes_top_ranked_archived_chunk_neighbors_returns_full_limit` passes vacuously.** It asserts `!entry.tags.iter().any(|t| t.starts_with("tier/"))` on entries whose tags were never hydrated. The assertion cannot fail. Instrumenting the fill loop shows all eight archived entries reaching the filter with `tags=[] passes_exclude=true` while the test reports ok.

## The fix, attached and verified

Applies cleanly with `git apply` to head `c28020d`. Two one-line query changes plus four tests that actually catch the regression.

Fail-then-pass, run from a worktree at `c28020d`:

```
env -u MX_SURREAL_MODE -u MX_SURREAL_URL -u MX_SURREAL_USER \
    -u MX_SURREAL_PASS_FILE -u MX_SURREAL_NS -u MX_SURREAL_DB \
    -u MX_MEMORY_PATH -u MX_MEMORY_BACKEND \
  cargo test --bins -- batch_tag_hydration batch_applicability_hydration \
                       semantic_search_exclude_tags_drops \
                       semantic_search_returns_chunked
```

Tests only, on clean head:

```
test result: FAILED. 0 passed; 4 failed; 0 ignored; 0 measured; 1193 filtered out
```

Tests plus the fix:

```
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 1193 filtered out
```

Full suite with the patch applied: 1194 unit tests plus the integration binaries, all green. `cargo fmt` and `cargo clippy` clean.

Full patch below.

Patch notes, verbatim (the local path in the apply line is where the file lives on kautau's box; the diff itself is below):

```text
Fix batch tag/applicability hydration in #401, plus the tests that catch it.

Apply to PR #401 (coryzibell/mx, branch feat/exclude-tags-search-list) at head
c28020d31135b176582642d3246c7def64c6d55f:

    git apply /home/kautau/.crewu/vol/patch-401-exclude-tags-hydration.diff

THE BUG
-------
Both batch hydration helpers #401 added filter with `WHERE in IN $knowledge`.
`in` is a SurrealQL keyword, so that clause never binds the edge's in-field and
matches nothing. Both call sites then swallow the empty result with
`response.take(0).unwrap_or_default()`, reporting it as "this entry has no
tags" rather than erroring.

  src/surreal_db/relationships.rs:296  get_tags_for_entries_async
  src/surreal_db/relationships.rs:378  get_applicability_for_entries_async

Measured on an in-memory store, same fixture, same bound Vec<Thing>:

  WHERE in IN $k         -> 0 rows
  WHERE $k CONTAINS in   -> 1 row  (correct)
  WHERE in INSIDE $k     -> 0 rows

IMPACT
------
get_knowledge_batch_async hydrates every semantic chunk candidate with tags=[],
so keep_after_exclude(&[], prefixes) is always true and NO chunked entry is
ever excluded. `mx memory search --semantic --exclude-tags` does nothing on the
chunked path -- the headline feature of #401. Unchunked entries are unaffected;
they filter in SQL via build_exclude_tags_filter, which does not read hydrated
tags.

Second effect: chunked entries come back from semantic search with tags and
applicability stripped, with or without any flag. The batch path replaced
per-entry get_knowledge_async, which hydrated both.

WHY THE EXISTING TESTS MISS IT
------------------------------
semantic_search_excludes_top_ranked_archived_chunk_neighbors_returns_full_limit
asserts `!entry.tags.iter().any(|t| t.starts_with("tier/"))` on entries whose
tags were never hydrated. The assertion cannot fail. Instrumenting the fill
loop shows all eight archived entries reaching the filter with
`tags=[] passes_exclude=true` while the test reports ok.

The four tests below assert on id sets and on cross-path agreement instead.

FAIL-THEN-PASS
--------------
Test file: src/surreal_db/tests.rs

  cd <worktree at c28020d>
  env -u MX_SURREAL_MODE -u MX_SURREAL_URL -u MX_SURREAL_USER \
      -u MX_SURREAL_PASS_FILE -u MX_SURREAL_NS -u MX_SURREAL_DB \
      -u MX_MEMORY_PATH -u MX_MEMORY_BACKEND \
    cargo test --bins -- batch_tag_hydration batch_applicability_hydration \
                         semantic_search_exclude_tags_drops \
                         semantic_search_returns_chunked

Tests only (before the relationships.rs hunk):
  test result: FAILED. 0 passed; 4 failed; 0 ignored; 0 measured; 1193 filtered out

Tests plus fix (this whole patch):
  test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 1193 filtered out

NOT FIXED HERE (separate, pre-existing defect in main)
------------------------------------------------------
get_applicability_for_entry_async (src/surreal_db/relationships.rs:328) runs
`SELECT VALUE meta::id(out)` -- a string -- and deserializes into Vec<Thing>:

  Serialization error: failed to deserialize; expected an object-like struct
  named $surrealdb::private::sql::Thing, found "backend"

unwrap_or_default() swallows that, so the single-entry applicability path
returns an empty vec for every entry. That predates #401 and is why the
applicability test below pins the batch path against the written edge rather
than against the single-entry path as an oracle.
```

```diff
diff --git a/src/surreal_db/relationships.rs b/src/surreal_db/relationships.rs
index 1e74660..91ab52c 100644
--- a/src/surreal_db/relationships.rs
+++ b/src/surreal_db/relationships.rs
@@ -295,8 +295,13 @@ impl SurrealDatabase {
 
         let mut response = with_db!(self, db, {
             db.query(
+                // `in` is a SurrealQL keyword: `in IN $knowledge` does not
+                // bind the edge's in-field and matches nothing, which
+                // `take(0).unwrap_or_default()` below then reports as "no
+                // tags". Comparing the other way round keeps `in` in field
+                // position.
                 "SELECT meta::id(in) AS entry_id, out.name AS tag \
-                 FROM tagged_with WHERE in IN $knowledge",
+                 FROM tagged_with WHERE $knowledge CONTAINS in",
             )
             .bind(("knowledge", things))
             .await
@@ -375,8 +380,9 @@ impl SurrealDatabase {
 
         let mut response = with_db!(self, db, {
             db.query(
+                // Same keyword collision as the tag query above.
                 "SELECT meta::id(in) AS entry_id, meta::id(out) AS applies_id \
-                 FROM applies_to WHERE in IN $knowledge",
+                 FROM applies_to WHERE $knowledge CONTAINS in",
             )
             .bind(("knowledge", things))
             .await
diff --git a/src/surreal_db/tests.rs b/src/surreal_db/tests.rs
index ad618c6..90beec1 100644
--- a/src/surreal_db/tests.rs
+++ b/src/surreal_db/tests.rs
@@ -3808,3 +3808,154 @@ fn semantic_search_no_flag_category_thinned_chunk_query_fills_to_limit() {
         );
     }
 }
+
+// =============================================================================
+// Batch hydration of tags / applicability (#401 regression)
+//
+// #401 replaced the per-entry `get_knowledge_async` hydration in the semantic
+// chunk fill loop with `get_knowledge_batch_async`, which fans out to
+// `get_tags_for_entries_async` and `get_applicability_for_entries_async`. Both
+// of those filter with `WHERE in IN $knowledge`. `in` is a SurrealQL keyword,
+// so that clause does not bind the edge's in-field and matches nothing; both
+// call sites then swallow the empty result with `take(0).unwrap_or_default()`
+// and report "this entry has no tags".
+//
+// Two consequences, both user-visible:
+//   1. `keep_after_exclude(&[], prefixes)` is always true, so `--exclude-tags`
+//      silently drops nothing on the chunked path -- the headline feature of
+//      #401. (Unchunked entries are unaffected; they filter in SQL.)
+//   2. Chunked entries come back from semantic search with their tags and
+//      applicability stripped, with or without any flag.
+//
+// The existing chunk-exclusion tests do not catch either, because they assert
+// `!entry.tags.iter().any(...)` on the same entries whose tags were never
+// hydrated -- an assertion that cannot fail. These tests assert on id sets and
+// on cross-path agreement instead.
+// =============================================================================
+
+#[test]
+fn batch_tag_hydration_agrees_with_the_single_entry_path() {
+    let db = SurrealDatabase::open_in_memory().unwrap();
+
+    let mut e = make_test_entry("kn-batch-tags", 5, 0.0);
+    e.content_hash = Some("hash-kn-batch-tags".to_string());
+    e.tags = vec!["tier/archived".to_string(), "project/x".to_string()];
+    db.upsert_knowledge(&e).unwrap();
+
+    let mut single = db.get_tags_for_entry("kn-batch-tags").unwrap();
+    single.sort();
+    assert_eq!(
+        single,
+        vec!["project/x".to_string(), "tier/archived".to_string()],
+        "precondition: the single-entry path must see the tags that were just \
+         written, otherwise this test proves nothing about the batch path"
+    );
+
+    let batched = SurrealDatabase::runtime()
+        .block_on(db.get_tags_for_entries_async(&["kn-batch-tags".to_string()]))
+        .unwrap();
+    let mut batch_tags = batched.get("kn-batch-tags").cloned().unwrap_or_default();
+    batch_tags.sort();
+
+    assert_eq!(
+        batch_tags, single,
+        "the batch tag query must return what the single-entry query returns; \
+         an empty result here means the WHERE clause matched nothing and every \
+         caller will read the entry as untagged"
+    );
+}
+
+#[test]
+fn batch_applicability_hydration_returns_the_edge_that_was_written() {
+    // NOTE: the single-entry path is NOT usable as the oracle here. It runs
+    // `SELECT VALUE meta::id(out)` -- a string -- and deserializes it into
+    // `Vec<Thing>`, which errors and is swallowed by `unwrap_or_default()`, so
+    // it returns an empty vec for every entry. That is a separate, pre-existing
+    // defect in `get_applicability_for_entry_async` and is NOT what this patch
+    // fixes. The `applies_to` edge itself is written correctly; a raw
+    // `SELECT meta::id(in), meta::id(out) FROM applies_to` returns it.
+    //
+    // What this test pins is the batch path, which projects `applies_id` as a
+    // String and types it correctly -- so once its WHERE clause actually
+    // matches, it returns the right value.
+    let db = SurrealDatabase::open_in_memory().unwrap();
+
+    let mut e = make_test_entry("kn-batch-app", 5, 0.0);
+    e.content_hash = Some("hash-kn-batch-app".to_string());
+    e.applicability = vec!["backend".to_string()];
+    db.upsert_knowledge(&e).unwrap();
+
+    let batched = SurrealDatabase::runtime()
+        .block_on(db.get_applicability_for_entries_async(&["kn-batch-app".to_string()]))
+        .unwrap();
+
+    assert_eq!(
+        batched.get("kn-batch-app").cloned().unwrap_or_default(),
+        vec!["backend".to_string()],
+        "the batch applicability query must return the applies_to edge that \
+         upsert_knowledge wrote; it shares the broken WHERE shape with the tag \
+         query and returns nothing for every entry"
+    );
+}
+
+#[test]
+fn semantic_search_exclude_tags_drops_a_chunked_entry_with_no_eligible_neighbors() {
+    use crate::store::{AgentContext, KnowledgeFilter};
+
+    // The narrowest possible statement of the feature: one archived, chunked
+    // entry, excluded by prefix, must not come back. No eligible entries are
+    // seeded, so the result set is either empty or wrong -- there is nothing
+    // for a vacuous tag assertion to hide behind.
+    let db = SurrealDatabase::open_in_memory().unwrap();
+    let mut query = vec![0.0f32; 8];
+    query[0] = 1.0;
+
+    seed_chunked_entry_with_tags(&db, "kn-only-archived", query.clone(), &["tier/archived"]);
+
+    let ctx = AgentContext::public_only();
+    let filter = KnowledgeFilter {
+        exclude_tag_prefixes: vec!["tier/".to_string()],
+        ..Default::default()
+    };
+    let results = db.semantic_search(&query, &ctx, &filter, 5).unwrap();
+
+    assert!(
+        results.is_empty(),
+        "--exclude-tags 'tier/' must drop a chunked entry tagged tier/archived; \
+         got {:?}",
+        results.iter().map(|e| e.id.as_str()).collect::<Vec<_>>()
+    );
+}
+
+#[test]
+fn semantic_search_returns_chunked_entries_with_their_tags_hydrated() {
+    use crate::store::{AgentContext, KnowledgeFilter};
+
+    // Independent of any filter: an entry reached through the embedding_chunk
+    // phase must carry its tags out. This is what makes the neighbouring
+    // exclusion tests meaningful -- an assertion of the form
+    // `!entry.tags.iter().any(...)` is worthless if tags always arrive empty.
+    let db = SurrealDatabase::open_in_memory().unwrap();
+    let mut query = vec![0.0f32; 8];
+    query[0] = 1.0;
+
+    seed_chunked_entry_with_tags(&db, "kn-tagged-chunk", query.clone(), &["project/x"]);
+
+    let ctx = AgentContext::public_only();
+    let filter = KnowledgeFilter::default();
+    let results = db.semantic_search(&query, &ctx, &filter, 5).unwrap();
+
+    assert_eq!(
+        results.len(),
+        1,
+        "precondition: the chunked entry must be found at all; got {:?}",
+        results.iter().map(|e| e.id.as_str()).collect::<Vec<_>>()
+    );
+    assert_eq!(
+        results[0].tags,
+        vec!["project/x".to_string()],
+        "a chunked entry must come back with its tags; empty tags here mean the \
+         batch hydration dropped them and every tag assertion downstream is \
+         vacuous"
+    );
+}
```

## Also asked, same area

Turn the two `response.take(0).unwrap_or_default()` calls on those query results into propagated errors. The whole reason this shipped through review, CI on three platforms, and a passing test asserting on the exact field it empties, is that a failing query returned "no tags" instead of an error. Make the next silent break loud.

## Non-blocking, noted for follow-up

Issues are being filed separately for these; none of them should hold this PR.

- **The fill-loop `have_enough` gate counts candidates instead of comparing scores.** `have_enough = scored_entries.len() >= limit` trips as soon as the unchunked phase has dumped `limit` entries in, including low-scoring ones, so the chunk scan terminates before reaching better-ranked chunked entries. At `--min-resonance 8 --limit 50` the resulting top-50 has 6 wrong keeps and 6 wrong drops, splitting exactly on `chunk_count`: every wrongly-kept entry is unchunked, every wrongly-dropped one is chunked. Same failure class as main's, much smaller blast radius. The fix is to gate the break on a score threshold — lowest kept score must beat the next unfetched chunk's score.
- **The cap-short-return warning fires on ordinary queries and names a cause that isn't there.** Repro with no `--exclude-tags` anywhere: `mx memory search --semantic "wake ritual" --category insight --limit 20` prints `likely excluded-tag dominance in the candidate set` while `exclude_tag_prefixes` is empty. Reproduced across `--category insight/decision/gotcha/technique/pattern` and `--min-resonance 9/10`. Most of those are correct, complete answers being warned about. Real trigger: category and resonance filters are applied to chunk candidates after hydration, starving `scored_entries` and driving the loop to the 1000-row floor, which sits just under the ~1507-row chunk table. It also has an off-by-one at exactly 1000 rows.
- **Draft PR #421** adds test coverage for that warning branch, stacked on this PR's head. It merges after this one.

## On the review record

kautau's Jul 8 approval was a plain issue comment, not a submitted review, so it never cleared the Jul 4 `CHANGES_REQUESTED`. This review supersedes that one and is the current state of the record.

Fix the hydration and this lands. Everything else in here is worth having.

gabaum10:
## We did not take your fix. It's correct, and it costs ~200x on the query plan — evidence below.

Everything else you asked for is in. Head is `9c57ad3`, rebased onto `main` @ `c2aa6330`.

### Why the supplied patch didn't ship

`WHERE $knowledge CONTAINS in` returns the right rows and plans as a full table scan. Measured with `EXPLAIN` against in-memory stores, 20k edges in `tagged_with`, 1000 bound ids:

| form | plan | index | rows | time |
|---|---|---|---|---|
| `WHERE $knowledge CONTAINS in` (your patch) | `Iterate Table` | none | 1000 | **9.2–11.5 s** |
| `WHERE in IN $knowledge` (the bug) | `Iterate Index` | `tagged_with_unique`, `operator: 'union'` | **0** | 4–6 ms |
| `WHERE in = $one` (single-entry path) | `Iterate Index` | `tagged_with_unique`, `operator: '='` | 1 | 0.7–2 ms |
| `SELECT ->tagged_with->tag.name FROM $knowledge` (shipped) | `Iterate Thing` ×1000 | n/a — direct key | 1000 | **82–107 ms** |

Cost on the `CONTAINS` form is O(table rows × bound array length):

| edges | bound ids | `CONTAINS` | traversal |
|---|---|---|---|
| 5,000 | 1000 | 2.26 s | 99 ms |
| 20,000 | 1000 | 9.2–11.5 s | 82–90 ms |
| 40,000 | 1000 | **18.9–24.4 s** | 99–107 ms |

Doubling either side doubles the query; the traversal form is flat in table size. At 40k edges the two differ by ~200x. Since these helpers are the batch hydration path, that lands directly on the axis this PR is sold on.

We shipped the traversal form instead. Plan verified by `EXPLAIN`, not inferred: `EXPLAIN FULL` at 40k edges / 1000 bound ids returns `Iterate Thing` for every one of the 2000 plan entries across both queries, with zero occurrences of `Iterate Table`. Measured run time at that scale: 92.4 ms for the tag query, 131.9 ms for applicability — flat in table size.

The applicability query needed one extra thing worth noting: `applicability_type` has no `name` field, its record id *is* its identity, so the traversal target is `.id`; and `meta::id()` does not vectorize over an array (it returns `NONE`), hence the explicit `array::map(…, |$v| meta::id($v))`.

### Your diagnosis is falsified, and the real mechanism is worth having

This was never a keyword collision. `in IN $knowledge` **binds correctly** — the plan carries all 1000 record ids resolved against the right field. It returns zero rows because the lookup becomes a `union` over the `in` prefix of the composite `tagged_with_unique` index on `(in, out)`, and a `union` over that prefix yields nothing, while `=` on the same prefix returns correctly. Adding a single-column `DEFINE INDEX … FIELDS in` does not rescue it — the planner still selects the composite and still emits `union`.

That also explains the thing your review couldn't: why `:229` `WHERE in = $fact` and `:254` `WHERE in = $knowledge` work fine while the batch form returns nothing. Same field, same position, different operator.

### Your suite structurally cannot catch this, and that's worth fixing

Every test fixture is small enough that `Iterate Table` and `Iterate Index` are indistinguishable — the scan wins on tiny data. So a 200x plan regression passes green on all three platforms, exactly the way the hydration bug did. It's the same failure class you caught, one layer down: an assertion that cannot fail, and now a benchmark that cannot notice.

### Part of the headline win is the bug

The 3988ms → 2738ms figure, and the 4x at `--limit 500`, were measured with the broken query in place — returning zero rows in 4–6 ms. Hydration was fast because it wasn't doing the work. We did not re-run your 67-query benchmark, so we can't hand you a corrected end-to-end figure and won't invent one. What we can say precisely: the hydration queries themselves now cost 92 ms (tags) and 132 ms (applicability) per 1000 entries, where the broken form cost 4–6 ms to return nothing. So the win shrinks by roughly that much and remains a clear win over the N+1 — but the published figure was measured against a path that wasn't doing the work, and the PR body shouldn't keep it as written. Your benchmark, your call on the re-run.

### The part that made this recoverable: your four tests

They assert on id sets and on cross-path agreement with the single-entry path — behavior, not implementation. That is the only reason we could replace the query form and still be held to your standard. They're unmodified, and they pass against the traversal form. A test that had asserted on the query text would have let this regression through while looking green; yours didn't.

### Dispositions

| Finding | Disposition | Evidence |
|---|---|---|
| **Blocker:** batch hydration never binds | **Fixed, different form** | Traversal instead of `CONTAINS`. Your four tests unmodified and passing; fail-then-pass confirmed the tests bind (4 failed pre-fix, 4 passed post-fix). |
| **Also asked:** propagate the two swallowed `take(0).unwrap_or_default()` | **Fixed as asked** | Both now `.context("…")?` in the file's existing `anyhow` idiom. A legitimately-empty result still returns `Ok(vec![])`; only genuine failures propagate. |
| Non-blocking: `have_enough` counts candidates not scores | **Deferred** | Per your note that issues are filed separately and shouldn't hold this. |
| Non-blocking: cap-short-return warning names a cause that isn't there | **Deferred** | Same. |
| Not in your review: PR was `CONFLICTING` against `main` | **Fixed** | `main` moved to `c2aa6330` about twenty minutes after you submitted, so "fix the hydration and this lands" predates the conflict. Rebased; one append/append hunk in `tests.rs`; `git diff --numstat origin/main..HEAD -- src/surreal_db/tests.rs` returns `659 0` — zero deletions against `main`, nothing of yours dropped. |
| Repo rule: CHANGELOG | **Added** | Your `CLAUDE.md` asks for a note when existing command behavior changes; `memory search`/`list` now return tags on chunked entries. |

Full suite on the new head: `1196 passed; 0 failed; 3 ignored` on the main binary, plus 1/12/10/10/22 on the integration binaries — matching the pre-change baseline exactly. `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean.

### Two things left alone, on purpose

**Three more `take(0).unwrap_or_default()` sites remain in the same file** — `:201` `get_facts_for_session_async`, `:229` `get_session_for_fact_async`, `:254` `get_tags_for_entry_async`. Same fail-silent class you asked to make loud. Left them because you scoped the ask to the two batch helpers, and because "unrelated changes inflating this diff" was your own earlier finding. One-liner each; say the word.

**#421** — untouched, and your draft is intact. To be precise about what the force-push did: `c28020d` still exists and is still in `test/cap-short-return-coverage`'s own history — you can check it out and it's all there. What changed is that it is no longer reachable from `feat/exclude-tags-search-list`, so your draft's base is no longer part of this PR's history and #421 will need a rebase onto the new head. Sorry for the churn; the conflict against `main` had to be cleared.

coryzibell (approved):
Tested all of it. You are right and I was wrong, twice.

`WHERE in = $one` returns the row on the same fixture where `in IN $knowledge` returns nothing, so the field binds and my keyword story is dead. `EXPLAIN FULL` on the patch I handed you returns `Iterate Table` over `tagged_with`, a full scan, exactly as you said. My sweep at a thousand bound ids: the scan runs 1,434 ms at five thousand edges, 5,858 at twenty thousand and 11,490 at forty thousand, while your traversal runs 264, 897 and 955. Doubling the table doubles the scan and moves your form by six percent. Your scaling law reproduces exactly. Your two hundred times does not on my hardware, I measure twelve, but the shape of the argument is right and the shape is what decides this.

Row equivalence held at all three scales. Your form returns precisely what mine would have returned, which is the thing that made this safe to swap.

Three corrections I owe you, on the record:

1. The `in` keyword diagnosis was wrong. Measured, not conceded.
2. The performance table in my review was measured against a query returning nothing, which it did in 4 ms. That is not a benchmark, it is an empty result timed. The table should not stand as written and I will re-run it. You were right not to invent a replacement.
3. The wrong diagnosis was still sitting in the header comment at `src/surreal_db/tests.rs`, contradicting your correct one in `relationships.rs`. That is my text and my job, not yours.

Poppy's four tests pass unmodified on your head and all four fail when I revert `relationships.rs`.

**What I did to this branch.** Rather than send you round the loop again, I rebased it here and force-pushed. Two conflicts, both from `main` moving under you:

- `src/handlers/memory.rs`, one adjacency hunk. `main` deleted the trigger-check block in the doors work, you deleted the two exclude helpers directly above it when you moved them to `helpers.rs`. Both deletions are correct, so both sides go, and the shared closing brace goes with them.
- `CHANGELOG.md`, both sides adding under the same heading. Kept both.

On top of that there is one extra commit from me, `test: correct exclude-tags header comment`, which replaces my keyword claim with your mechanism: the bind is correct and reaches the in-field, and what fails is `in IN $knowledge` planning as a `union` lookup over the `in` prefix of the UNIQUE composite index on `(in, out)`. It cites the schema lines. It is the top commit and it is yours to drop if you would rather write that sentence yourself. Nothing else of yours was touched.

Rebased tree is green: 1226 unit including the new doors suite, 1303 across every binary, `cargo fmt` and `clippy -D warnings` clean.

Yes on the three remaining `unwrap_or_default` sites, in a separate pull request rather than this one.

